PR #92, opened 2021-01-22
Add SDK log config

matthewpaul:
This commit adds support for configuring the SDK logging. It provides configuration options on the CLI and via JSON, provides help information output from the executable, adds tests for this functionality, and adds an entry to the CHANGELOG describing the change. I have also modified existing log config variables to deconflict the naming and reduce confusion.


source/SharedCrtResourceManager.cpp
```diff
@@ -88,11 +89,23 @@ void SharedCrtResourceManager::initializeAllocator()
...
 
-int SharedCrtResourceManager::buildClient()
+int SharedCrtResourceManager::buildClient(const PlainConfig &config)
 {
     // We MUST declare an instance of the ApiHandle to perform global initialization
     // of the SDK libraries
     apiHandle = unique_ptr<ApiHandle>(new ApiHandle());
+    if (config.logConfig.sdkLoggingEnabled)
+    {
+        apiHandle->InitializeLogging(config.logConfig.sdkLogLevel, config.logConfig.sdkLogFile.c_str());
```
HarshGandhi-AWS, line 99: I think we should add permissions on this file. Or should ask user to add permissions and we should validate those permissions on DC side

source/config/Config.cpp
```diff
@@ -296,13 +304,49 @@ int PlainConfig::LogConfig::ParseLogLevel(string level)
...
+            "Provided log level %s is not a known log level for the AWS IoT Device Client", Sanitize(level).c_str()));
+    }
+}
+
+Aws::Crt::LogLevel PlainConfig::LogConfig::ParseSDKLogLevel(string level)
+{
+    string temp = level;
+    std::transform(temp.begin(), temp.end(), temp.begin(), [](unsigned char c) { return std::toupper(c); });
+
+    if ("TRACE" == temp)
```
lwilkovich, line 318: This is fine, but a switch might be better

source/config/Config.h
```diff
@@ -102,9 +104,20 @@ namespace Aws
...
+                    static constexpr char JSON_KEY_ENABLE_SDK_LOGGING[] = "enable-sdk-logging";
+                    static constexpr char JSON_KEY_SDK_LOG_LEVEL[] = "sdk-log-level";
+                    static constexpr char JSON_KEY_SDK_LOG_FILE[] = "sdk-log-file";
+
+                    int deviceClientlogLevel{3};
+                    std::string deviceClientLogtype;
+                    std::string deviceClientLogFile;
+                    bool sdkLoggingEnabled = false;
+                    Aws::Crt::LogLevel sdkLogLevel = Aws::Crt::LogLevel::Trace;
+                    std::string sdkLogFile = "/var/log/aws-iot-device-client/sdk.log";
```
HarshGandhi-AWS, line 120: Create a Default_Path constant over [here](https://github.com/awslabs/aws-iot-device-client/blob/main/source/config/Config.h#L220-L225) to save path and use it in DC application code and in unit tests.

source/SharedCrtResourceManager.cpp
matthewpaul: After speaking with our AppSec team and the SDK team, there aren't currently any concerns around securing/validating the permissions of the SDK log file

source/config/Config.h
matthewpaul: That's a good idea! I have created the constant, but rather than placing it at the region of Config.h that you specified, I placed it in the PlainConfig::LogConfig namespace

lwilkovich (approved):

source/config/Config.h
HarshGandhi-AWS: No problem. We should consider moving other defaults in respective "PlainConfig::.." namespaces. But that change is for a different PR.

HarshGandhi-AWS (approved):
LGTM

PR #95, opened 2021-01-25
Doxygen actions update

lwilkovich:
*Description of changes:*
Doxygen actions update when merging main


.github/workflows/ci.yml
```diff
@@ -30,6 +30,34 @@ jobs:
...
+        cd docs
+        doxygen
+        cp -r ./html/* ./
+        rm -rf ./html
+        git add .
+    - name: Commit files
+      run: |
+        git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        git config --local user.name "github-actions[bot]"
+        git commit -m "Add changes" -a
```
matthewpaul, line 53: Can we make this commit message a little more descriptive? Something like "Auto-generate Doxygen documentation via Github Actions" would probably be sufficient

.github/workflows/ci.yml
```diff
@@ -30,6 +30,34 @@ jobs:
...
+    - name: Push changes
+      uses: ad-m/github-push-action@master
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
```
matthewpaul, line 57: I thought we had discussed not using a Github secret? Do we need to do this here if the branch is not protected?

lwilkovich: This is not something I had to make, it is automatically created to authenticate a workflow.  We use it in the other ci too.

HarshGandhi-AWS (approved):

matthewpaul (approved):

PR #99, opened 2021-01-26
Small cleanup in DD & making FP more consistent with the other features

lwilkovich:
*Description of changes:*
Making FP more consistent with the other features.  This will make it look less confusing in doxygen.  Also removing some redundant things from DD


matthewpaul (approved):

source/main.cpp
```diff
@@ -39,6 +39,9 @@ using namespace Aws::Iot::DeviceClient::Logging;
 #if !defined(EXCLUDE_ST)
 using namespace Aws::Iot::DeviceClient::SecureTunneling;
 #endif
+#if !defined(EXCLUDE_FP)
+using namespace Aws::Iot::DeviceClient::FleetProvisioningNS;
```
HarshGandhi-AWS, line 43: Can you update the namespace to FleetProvisioning and Class name to FleetProvisioningFeature to maintain naming standards across the program ?

lwilkovich: We don't want to change the class name to include feature because it doesn't implement the same feature interface.

HarshGandhi-AWS (approved):

PR #102, opened 2021-01-27
Add abort if MQTT client configuration fails

matthewpaul:
This commit addresses an issue where we may continue with feature startup even if we fail to build the MQTT client. The issue was caused by bubbling up the return code of an SDK function when we should just be returning the ABORT constant.

### DIFF
--- a/source/SharedCrtResourceManager.cpp
+++ b/source/SharedCrtResourceManager.cpp
@@ -157,7 +157,7 @@ int SharedCrtResourceManager::establishConnection(const PlainConfig &config)
             TAG,
             "MQTT Client Configuration initialization failed with error: %s",
             ErrorDebugString(clientConfig.LastError()));
-        return clientConfig.LastError();
+        return ABORT;
     }
 
     connection = mqttClient->NewConnection(clientConfig);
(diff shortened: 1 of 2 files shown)


lwilkovich (approved):

source/SharedCrtResourceManager.cpp
```diff
@@ -211,7 +211,7 @@ int SharedCrtResourceManager::establishConnection(const PlainConfig &config)
     if (!connection->Connect(config.thingName->c_str(), true, 0))
     {
         LOGM_ERROR(TAG, "MQTT Connection failed with error: %s", ErrorDebugString(connection->LastError()));
-        return connection->LastError();
+        return RETRY;
```
HarshGandhi-AWS, line 214: Why do we want to retry establishing connection over here?

matthewpaul: At this point we are not dealing with a client configuration error, we're dealing with a problem connecting to IoT Core. This could be caused by network adapters, network issues, API availability, or improperly configured IoT policies/certificates. Since these issues are all external to the Device Client, we attempt retry in the hope that the problem will be mitigated.

HarshGandhi-AWS (approved):

janborch:
I confirm that the patch solves the problem, I get now an ABORT and meaningful message in my logs. Thanks for the quick fix! 

```
Main.cpp}: *** AWS IOT DEVICE CLIENT FATAL ERROR: Failed to establish the MQTT Client. Please verify your AWS IoT credentials, configuration and/or cer
tificate policy. ***^@

```

Issue #101, opened 2021-01-27
When using Fleet provisioning with a CSR the runtime config points to the wrong private key (bootstrap cert key)


janborch:
I use the following device client config file, where I point to my bootstrapping cert and key files in the main section. In the fleet provisioning section I reference a CSR I generated from a newly generated device key.

```json
{
  "endpoint": "xxxxxxxxxx-ats.iot.eu-west-1.amazonaws.com",
  "cert": "/home/ubuntu/adc/bootstrap.pem",
  "key": "/home/ubuntu/adc/bootstrap.key",
  "root-ca": "/home/ubuntu/adc/AmazonRootCA1.pem",
  "thing-name": "UbuntuEC201",
  "logging": {
    "level": "DEBUG",
    "type": "FILE",
    "file": "/var/log/aws-iot-device-client/aws-iot-device-client.log"
  },
  "jobs": {
    "enabled": true,
    "handler-directory": "/home/ubuntu/adc/handler"
  },
  "tunneling": {
    "enabled": true
  },
  "device-defender": {
    "enabled": true,
    "interval": 60
  },
  "fleet-provisioning": {
    "enabled": true,
    "template-name": "ProvisioningTemplate-bad1wpiFF6M0",
    "csr-file": "/home/ubuntu/adc/deviceec201.csr"
  }
}
```

The provisioning succeeds and the following runtime config is generated, but the `key` entry points to the wrong key. It should point to my device key not the bootstrap key.
```json
{
"runtime-config": {
    "completed-fp": true,
    "cert": "/root/.aws-iot-device-client/keys/d65e7933fd7b363071a737d4be24d28d8f3614d9aea17f6dd72839ac2c1d560b-certificate.pem.crt",
    "key": "/home/ubuntu/adc/bootstrap.key",
    "thing-name": "ThingFP-d65e7933fd7b363071a737d4be24d28d8f3614d9aea17f6dd72839ac2c1d560b"
    }
}
```
Additionally when you restart the client with this wrong config the client terminates with a segmentation fault.

## Expected runtime config:

I should be able to provide my device key location along the device CSR location in the config file 

```json
...
  "fleet-provisioning": {
    "enabled": true,
    "template-name": "ProvisioningTemplate-bad1wpiFF6M0",
    "csr-file": "/home/ubuntu/adc/deviceec201.csr",
    "key": "/home/ubuntu/adc/deviceec201.key"
  }
```

[...]

PR #104, opened 2021-02-02
Added support for new config option for taking device private key location

HarshGandhi-AWS:
*Issue #101*

*Description of changes:*
With this commit, I am adding config option for user to provide the device private key location via CLI or JSON file. This option will be required for provisioning the device using CSR file (CreateCertificateFromCsr API). If the device key or the CSR file is not provided, Device Client will use claim certificate and bootstrap private key (CreateKeysAndCertificate API) for provisioning the device.

### DIFF
--- a/README.md
+++ b/README.md
@@ -558,9 +558,9 @@ More details about AWS IoT Fleet Provisioning by claim can be found here: https:
 
 *Note: If the fleet provisioning feature fails to provision the new key, certificate or thing/device, the device client will abort with fatal error.*
 
-*Note: If the CSR file or the Device Private Key is not provided, the Device Client will use Claim certificate and Private key for provisioning the device*
+*Note: If the CSR file is specified without also specifying a device private key, the Device Client will use Claim Certificate and Private key to generate new Certificate and Private Key while provisioning the device*
 
-*Note: Please make sure that the Claim certificate, private key and/or CSR file, device private key stored on device side have respective permissions applied to it as mentioned above in the "File and Directory Permissions" section of the readme.*
+*Note: Please make sure that the Claim certificate, private key, CSR file and/or device private key stored on device side have respective permissions applied to it as mentioned above in the "File and Directory Permissions" section of the readme.*
 
 Refer to the [CreateKeysAndCertificate](https://docs.aws.amazon.com/iot/latest/developerguide/fleet-provision-api.html#create-keys-cert) and [CreateCertificateFromCsr](https://docs.aws.amazon.com/iot/latest/developerguide/fleet-provision-api.html#create-cert-csr) APIs for more details.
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -558,9 +558,9 @@ More details about AWS IoT Fleet Provisioning by claim can be found here: https:
 
 *Note: If the fleet provisioning feature fails to provision the new key, certificate or thing/device, the device client will abort with fatal error.*
 
-*Note: If a CSR file is not provided, the Device Client will use Claim Certificate and Private key for provisioning the device.*
+*Note: If the CSR file or the Device Private Key is not provided, the Device Client will use Claim certificate and Private key for provisioning the device*
```
matthewpaul, line 561: Shouldn't this say - "if the CSR file is specified without also specifying a private key, the Device Client will..."

README.md
```diff
@@ -558,9 +558,9 @@ More details about AWS IoT Fleet Provisioning by claim can be found here: https:
...
 
-*Note: Please make sure that the Claim certificate, private key and/or CSR file stored on device side have respective permissions applied to it as mentioned above in the "File and Directory Permissions" section of the readme.*
+*Note: Please make sure that the Claim certificate, private key and/or CSR file, device private key stored on device side have respective permissions applied to it as mentioned above in the "File and Directory Permissions" section of the readme.*
```
matthewpaul, line 563: the and/or should be moved to right before "device private key"

matthewpaul (commented):
Logic looks good to me, might just need a few updates on the documentation language. Also, do we need to update:

1. The setup script
2. The configuration template file
3. Test cases

source/fleetprovisioning/FleetProvisioning.cpp
```diff
@@ -551,6 +555,34 @@ bool FleetProvisioning::ProvisionDevice(shared_ptr<SharedCrtResourceManager> fpC
...
+{
+    struct stat info;
+    bool locatedDeviceKey = true;
+    string expandedPath = FileUtils::ExtractExpandedPath(filePath.c_str());
+    if (stat(expandedPath.c_str(), &info) != 0)
+    {
+        LOGM_ERROR(
+            TAG,
+            "*** %s: Failed to find the device private key %s, file does not exist ***",
+            DeviceClient::DC_FATAL_ERROR,
```
matthewpaul, line 568: Is this actually a fatal error? I thought that if we can't find the private key, we are supposed to default back to the other FP API?

HarshGandhi-AWS: My bad. I'll update the log.

test/config/TestConfig.cpp
```diff
@@ -33,6 +33,12 @@ TEST(Config, AllFeaturesEnabled)
     "device-defender": {
         "enabled": true,
         "interval": 300
+    },
+    "fleet-provisioning": {
+        "enabled": true,
+        "template-name": "template-name",
+		"csr-file": "csr-file",
```
matthewpaul, line 40: It looks like the formatting here is a bit weird, csr-file and device-key should probably be aligned with enabled and template-name

matthewpaul (approved):

test/config/TestConfig.cpp
HarshGandhi-AWS: Looks like an tabs vs space issue (looks perfect in text editor). The change is required in every test case and will have a separate PR to resolve the issue.

lwilkovich (approved):

PR #114, opened 2021-02-11
Github Actions for building with shared libs && reorganized CI order

lwilkovich:
*Description of changes:*
Added actions for building artifacts with the shared_libs turned on.
Reordered the actions.
This also required some small changes to the build script.
Passing tests can be seen in the actions tab


.github/build.sh
```diff
@@ -111,6 +127,9 @@ case $compileMode in
       # Fix for the Cmake executing build of the sdk which errors out linking incorrectly to openssl
       cmake -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_TOOLCHAIN_FILE=../cmake-toolchain/Toolchain-armhf.cmake -DBUILD_SDK=ON -DEXCLUDE_JOBS=ON -DEXCLUDE_DD=ON -DEXCLUDE_FP=ON -DDISABLE_MQTT=ON ../ || true
       cmake -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_TOOLCHAIN_FILE=../cmake-toolchain/Toolchain-armhf.cmake -DBUILD_SDK=ON -DEXCLUDE_JOBS=ON -DEXCLUDE_DD=ON -DEXCLUDE_FP=ON -DDISABLE_MQTT=ON ../
+    elif [ "$sharedLibs" = true ]; then
+      cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_TOOLCHAIN_FILE=../cmake-toolchain/Toolchain-armhf.cmake -DBUILD_SDK=ON ../ || true
```
matthewpaul, line 131: Can you file a ticket for addressing the duplicate cmake calls in this file?

.github/workflows/ci.yml
```diff
@@ -59,9 +37,97 @@ jobs:
...
+      - name: Build ${{ env.PACKAGE_NAME }}
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u awslabs --password-stdin
+          export DOCKER_IMAGE=docker.pkg.github.com/awslabs/aws-iot-device-client/ubuntu-16-x64:latest
+          docker pull $DOCKER_IMAGE
+          docker run --mount type=bind,source=$(pwd),target=/root/${{ env.PACKAGE_NAME }} --env GITHUB_REF $DOCKER_IMAGE --compiler=clang-${{ matrix.version }}
+
   build-armhf32:
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/main'
```
matthewpaul, line 64: Why are you removing this conditional check? Won't this increase the status check times for all branches? And is there an account level limit for number of actions workflows we can run?

.github/workflows/ci.yml
```diff
@@ -137,11 +201,33 @@ jobs:
...
+      - name: Archive ubuntu-16-x64 Build Artifact                                                                                                                                                        
+        uses: actions/upload-artifact@v2                                                                                                                                                                      
+        # Run for main branch only                                                                                                                                                                            
+        if: github.ref == 'refs/heads/main'                                                                                                                                                                   
+        with:                                                                                                                                                                                                 
+          name: "SL.ubuntu.x64.${{ env.PACKAGE_NAME }}"
```
matthewpaul, line 212: I think this naming convention is confusing - the static builds get DC for Device Client but the shared library builds get SL for shared library. I think all of them should get DC to differentiate between ST builds and then somewhere in the artifact name you can just append "shared", with static builds just not having the extension, or having "static"

.github/workflows/ci.yml
lwilkovich: There is an account level limit for the number of workflows at a given time I believe.  And the number of jobs in a workflow is limited by time, where we're no where near the cap.  I also felt that we should be building these each time for more transparency, it doesn't really add much time to wait.

matthewpaul (approved):

HarshGandhi-AWS (approved):

PR #118, opened 2021-02-25
Added notice about charges to README, moved aws_http_library_init to sharedCRTResourceManager

lwilkovich:
*Description of changes:*
* Added Notice about charges incurred to Readme
* Moved aws_http_library_init to sharedCRTResourceManager

There is on reason to call aws_http_library_init if multiple features
require this especially when dealing with feature compilation.
It should be handled by the sharedCRTResourceManager where we can ignore
additional calls to initialize it.

### DIFF
--- a/source/SharedCrtResourceManager.cpp
+++ b/source/SharedCrtResourceManager.cpp
@@ -136,6 +136,22 @@ int SharedCrtResourceManager::buildClient(const PlainConfig &config)
     return SharedCrtResourceManager::SUCCESS;
 }
 
+void SharedCrtResourceManager::initializeAWSHttpLib()
+{
+    if (!initialized)
+    {
+        LOG_WARN(TAG, "Tried to aws_http_library_init but the SharedCrtResourceManager has not yet been initialized!");
+        return;
+    }
+    if (initializedAWSHttpLib)
+    {
+        LOG_WARN(TAG, "Tried to aws_http_library_init but it was already initialized!");
+        return;
+    }
+    aws_http_library_init(getAllocator());
+    initializedAWSHttpLib = true;
+}
+
 int SharedCrtResourceManager::establishConnection(const PlainConfig &config)
 {
     if (!locateCredentials(config))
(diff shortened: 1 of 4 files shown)


source/SharedCrtResourceManager.cpp
```diff
@@ -136,6 +136,21 @@ int SharedCrtResourceManager::buildClient(const PlainConfig &config)
     return SharedCrtResourceManager::SUCCESS;
 }
 
+int SharedCrtResourceManager::initializeAWSHttpLib() {
+    if (!initialized)
+    {
+        LOG_WARN(TAG, "Tried to aws_http_library_init but the SharedCrtResourceManager has not yet been initialized!");
+        return -1;
```
matthewpaul, line 144: What's the purpose of returning a value in here if you never check it?

matthewpaul (commented):
I think this change would benefit from having the motivation for the change within the commit message, since it's currently unclear why you needed to move the HTTP init to its own function/flag. Was there an issue against the repo that this is addressing?

source/SharedCrtResourceManager.cpp
```diff
@@ -136,6 +136,20 @@ int SharedCrtResourceManager::buildClient(const PlainConfig &config)
...
+    if (!initialized)
+    {
+        LOG_WARN(TAG, "Tried to aws_http_library_init but the SharedCrtResourceManager has not yet been initialized!");
+    }
+    if (initializedAWSHttpLib)
+    {
+        LOG_WARN(TAG, "Tried to aws_http_library_init but it was already initialized!");
+    }
+    aws_http_library_init(getAllocator());
+    initializedAWSHttpLib = true;
```
matthewpaul, line 150: I think you actually need a lock for this value, or make it atomic. I can imagine a situation where you have multiple features initializing at the same time and you're reading/writing to this value concurrently. Likely have a race condition here

matthewpaul (approved):

MiaoZhangAWS (approved):

PR #120, opened 2021-03-03
Improve Jobs feature shutdown stability

matthewpaul:
This commit addresses in an issue in which the jobs feature would not update a job execution successfully upon job completion if a shutdown request had been delivered to the jobs feature by main.cpp. The issue was resolved by adding support for passing an onSuccess callback function to our exponential backoff retry so that we only shut down the feature once we've made some attempts to update the job execution and have tried to wait for a response from the Jobs service. This commit also improves stability by replacing a group of booleans protected by a lock with atomics.


source/util/Retry.cpp
```diff
@@ -10,38 +10,58 @@ using namespace Aws::Iot::DeviceClient::Util;
...
+    bool needToStop = false;
+    if (config.needStopFlag != nullptr)
+    {
+        needToStop = config.needStopFlag->load();
+    }
+
+    if (needToStop)
+    {
+        LOG_DEBUG(TAG, "Stop flag was set prior to executing retryable function, will not attempt retryable execution");
+        return false;
```
lwilkovich, line 27: Should onComplete be called here?

source/util/Retry.h
```diff
@@ -61,11 +57,14 @@ namespace Aws
...
                      * whether it is successful or not, since this indicator is what will determine whether the
                      * function is retried or not.
+                     * @param onComplete a callback function which will be executed once this function is finished
+                     * attempting retries
                      * @param config the ExponentialRetryConfig specifying whether the function should be retried
                      * @return a bool representing whether the retryableFunction was successful or not
                      */
                     static bool exponentialBackoff(
                         std::function<bool()> retryableFunction,
+                        std::function<void()> onComplete,
```
lwilkovich, line 67: Would it be worth adding this function ptr to the ExponentialRetryConfig instead of passing it in this way?  That way we can have nullptr default by default and a user doesn't need to think about passing something in.

Could also change that function ->static bool exponentialBackoff(std::function<bool()> retryableFunction, ExponentialRetryConfig config, std::function<void()> onComplete
 = nullptr); making it optional

source/util/Retry.cpp
matthewpaul: I think that's a good idea! I will add it

source/util/Retry.h
matthewpaul: I think I like your second approach better, as it will maintain some consistency with the method signatures (IE I won't have one callback function as a parameter to the function and another as a parameter to the config struct.

source/jobs/JobsFeature.cpp
```diff
@@ -167,10 +173,19 @@ void JobsFeature::subscribeToUpdateJobExecutionStatusAccepted(string jobId)
         std::bind(
             &JobsFeature::updateJobExecutionStatusAcceptedHandler, this, std::placeholders::_1, std::placeholders::_2),
         std::bind(&JobsFeature::ackSubscribeToUpdateJobExecutionAccepted, this, std::placeholders::_1));
+
+    if (std::future_status::timeout == updateAcceptedPromise.get_future().wait_for(std::chrono::seconds(10)))
```
KareemAliAWS, line 177: Can't you use the promise returned by `SubscribeToUpdateJobExecutionAccepted` instead?

https://aws.github.io/aws-iot-device-sdk-js-v2/classes/aws_iot_device_sdk.iotjobsclient.html#subscribetoupdatejobexecutionaccepted

source/jobs/JobsFeature.cpp
```diff
@@ -180,6 +195,14 @@ void JobsFeature::subscribeToUpdateJobExecutionStatusRejected(string jobId)
         std::bind(
             &JobsFeature::updateJobExecutionStatusRejectedHandler, this, std::placeholders::_1, std::placeholders::_2),
         std::bind(&JobsFeature::ackSubscribeToUpdateJobExecutionRejected, this, std::placeholders::_1));
+
+    if (std::future_status::timeout == updateRejectedPromise.get_future().wait_for(std::chrono::seconds(10)))
```
KareemAliAWS, line 199: Can't you use the promise returned by subscribeToUpdateJobExecutionRejected instead?

https://aws.github.io/aws-iot-device-sdk-js-v2/classes/aws_iot_device_sdk.iotjobsclient.html#subscribetoupdatejobexecutionrejected

KareemAliAWS (changes requested):

source/jobs/JobsFeature.cpp
KareemAliAWS, line 177: So it seems that the purpose of the promise is to block until the SUBACK message is received and fail if it takes too long but I'm not sure whether `SubscribeToUpdateJobExecutionAccepted` blocks until it receives the SUBACK or not? Because if it blocks, then that promise won't be helpful. Can you refer me to the docs for the behavior of the SubscribeToUpdateJobExecutionAccepted?

I was only able to find the JS docs which does return a promise that resolves when the SUBACK is received but I don't know if the CPP SDK behaves similarly or not.

matthewpaul: I have no reason to believe that `SubscribeToUpdateJobExecutionAccepted` will actually block until the suback is received. This would be the main motivator for having the sub ack in the first place. The updateAcceptedPromise is set by the `ackSubscribeToUpdateJobExecutionAccepted` function. Therefore, this current code will block on line 177 while waiting for the promise's value to be set.

KareemAliAWS (dismissed):

lwilkovich (approved):

source/jobs/JobsFeature.cpp
```diff
@@ -93,6 +93,7 @@ void JobsFeature::ackSubscribeToUpdateJobExecutionAccepted(int ioError)
         LOG_ERROR(TAG, errorMessage.c_str());
```
HarshGandhi-AWS, line 93: We might wanna update the error message here and in other call back methods since it is same everywhere. (Line # 46, 63, 75, 92). User will not be able to debug which subscription call failed.

source/jobs/JobsFeature.cpp
```diff
@@ -167,10 +173,19 @@ void JobsFeature::subscribeToUpdateJobExecutionStatusAccepted(string jobId)
         std::bind(
             &JobsFeature::updateJobExecutionStatusAcceptedHandler, this, std::placeholders::_1, std::placeholders::_2),
         std::bind(&JobsFeature::ackSubscribeToUpdateJobExecutionAccepted, this, std::placeholders::_1));
+
+    if (std::future_status::timeout == updateAcceptedPromise.get_future().wait_for(std::chrono::seconds(10)))
+    {
+        ostringstream errorMessage;
+        errorMessage << "Failed to subscribe to updateJobExecutionStatusAccepted!";
```
HarshGandhi-AWS, line 180: We should update this error message from Failed to timed out. Here the subscription call has timed out.

source/jobs/JobsFeature.cpp
```diff
@@ -180,6 +195,14 @@ void JobsFeature::subscribeToUpdateJobExecutionStatusRejected(string jobId)
         std::bind(
             &JobsFeature::updateJobExecutionStatusRejectedHandler, this, std::placeholders::_1, std::placeholders::_2),
         std::bind(&JobsFeature::ackSubscribeToUpdateJobExecutionRejected, this, std::placeholders::_1));
+
+    if (std::future_status::timeout == updateRejectedPromise.get_future().wait_for(std::chrono::seconds(10)))
+    {
+        ostringstream errorMessage;
+        errorMessage << "Failed to subscribe to updateJobExecutionStatusRejected!";
```
HarshGandhi-AWS, line 202: Update the error messages everywhere. We should have timed out message here and failure message in respective call back methods.

HarshGandhi-AWS (changes requested):

source/jobs/JobsFeature.cpp
matthewpaul: That's a good suggestion, I'll make the update

source/jobs/JobsFeature.cpp
matthewpaul: I think the fact that it timed out indicates that it has failed, but I agree that it makes it a bit clearer for debugging so I will update it

source/jobs/JobsFeature.cpp
matthewpaul: Agree, will update

source/jobs/JobsFeature.cpp
```diff
@@ -43,9 +43,9 @@ void JobsFeature::ackSubscribeToNextJobChanged(int ioError)
     if (ioError)
     {
         // TODO We need to implement a strategy for what do when our subscription fails
-        string errorMessage = "Failed to subscribe to nextJobChanged";
+        string errorMessage = "Encountered an ioError while attempting to subscribe to NextJobChanged";
         LOG_ERROR(TAG, errorMessage.c_str());
```
HarshGandhi-AWS, line 47: Also consider printing the error coming from IoT core. It will help user to debug the issue more easily. You can use "ErrorDebugString" method present in CRT SDK for this.

matthewpaul: I added logging of the specific io error in my latest revision, but I don't agree that we can use "ErrorDebugString" provided by the CRT SDK. I believe that method is reserved for printing readable versions of CRT SDK errors, but I'm pretty sure it won't make sense to have it try to interpret an error passed down from the Jobs service and/or the message broker.

HarshGandhi-AWS: I think we can use it. I have seen it is being used in various [sample](https://github.com/aws/aws-iot-device-sdk-cpp-v2/blob/main/samples/shadow/shadow_sync/main.cpp#L90) codes.

HarshGandhi-AWS (approved):
Looks good to me.

KareemAliAWS (approved):

PR #123, opened 2021-03-12
Update job to "In Progress" after receive

matthewpaul:
This commit sets the job execution status to in progress once the job is received and processable by the Device Client. In order to accomplish this, the Job engine code was moved to a separate thread, since its activity is typically long running and will block the main thread from receiving the update job execution response.


source/jobs/JobsFeature.cpp
```diff
@@ -534,121 +528,137 @@ bool JobsFeature::isDuplicateNotification(JobExecutionData job)
...
+}
 
-    LOGM_DEBUG(
-        TAG, "About to execute: %s %s", Sanitize(command.str()).c_str(), Sanitize(argsStringForLogging.str()).c_str());
-    unique_ptr<JobEngine> engine(new JobEngine);
-    int executionStatus = engine->exec_cmd(command.str().c_str(), args);
+void JobsFeature::executeJob(JobExecutionData job)
+{
+    LOGM_INFO(TAG, "Executing job: %s", job.JobId->c_str());
+    publishUpdateJobExecutionStatus(job, {Iotjobs::JobStatus::IN_PROGRESS, "", "", ""});
```
KareemAliAWS, line 574: What if the this API call fails? shouldn't the body of the `runJob` lambda be part of the onComplete callback for `publishUpdateJobExecutionStatus`?

source/jobs/JobsFeature.cpp
```diff
@@ -534,121 +528,137 @@ bool JobsFeature::isDuplicateNotification(JobExecutionData job)
...
+            TAG, "Did not find any arguments in the incoming job document. Value should be a JSON array of arguments");
     }
-    else
+
+    auto shutdownHandler = [this]() -> void {
+        handlingJob.store(false);
+        if (needStop.load())
+        {
+            LOGM_INFO(TAG, "Shutting down %s now that job execution is complete", getName().c_str());
+            baseNotifier->onEvent((Feature *)this, ClientBaseEventNotification::FEATURE_STOPPED);
```
KareemAliAWS, line 602: I think this might lead to a situation where the baseNotifier fires the the ClientBaseEventNotification::FEATURE_STOPPED if JobsFeature::stop where to execute in parallel with this method, although I'm not sure if thread model allows parallel execution of those two methods or not.

source/jobs/JobsFeature.h
```diff
@@ -268,6 +269,17 @@ namespace Aws
...
+                    /**
+                     * \brief Builds the command that will be executed
+                     * @param path the provided path to the executable
+                     * @param operation the name of the executable
+                     * @return the full executable path.
+                     *
+                     * If this command is unable to find a given job handler and/or the permissions
+                     * for the given job handler are inappropriate, this function will thrown an exception.
+                     */
+                    std::string buildCommand(Aws::Crt::String path, Aws::Crt::String operation);
```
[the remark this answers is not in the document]
matthewpaul: In this particular instance, Aws::Crt::String is the data type that's passed to the job handler that we register with the SDK. I'm not quite sure as to what the performance benefits/tradeoffs are, but I typically stick to using the std::string implementation unless it's absolutely required to use something else.

source/jobs/JobsFeature.cpp
matthewpaul: In this instance, the publishUpdateJobExecutionStatus method performs an exponential backoff while attempting to update the status to inProgress. I didn't want to make execution of the job dependent on being able to set the job status to in progress, since it's not a hard requirement for us to update the execution status to in progress in order to eventually reach a terminal state for the job execution.

source/jobs/JobsFeature.cpp
matthewpaul: I think it may technically be possible, but there are some thread safety mechanisms on the other side within the event handler (`handle_feature_stopped`) that increase thread safety. I can't currently identify any bad consequences if the event notifier were to fire twice, other than some wasted CPU cycles

KareemAliAWS (approved):

source/jobs/JobsFeature.cpp
KareemAliAWS: I'm not clear on would constitute a hard requirement but in our docs for jobs for the device workflow, we require the device to update the job execution status before executing the job action

https://docs.aws.amazon.com/iot/latest/developerguide/jobs-devices.html#jobs-workflow-device-online

HarshGandhi-AWS (approved):

PR #124, opened 2021-03-22
Seperating more advanced/technical information from the main readme.

lwilkovich:
*Description of changes:*
Reducing the 'wall of text'.  Several new readmes have been created to house the more advanced or technical information; these are all linked to from the main readme.

This might be easier to review just looking at the branch first & then at the diff.


satheshantanu (approved):

xps506 (dismissed):

source/jobs/README.md
```diff
@@ -0,0 +1,232 @@
...
+    + [Jobs Feature Configuration Options](#jobs-feature-configuration-options)
+      - [Configuring the Jobs feature via the command line](#configuring-the-jobs-feature-via-the-command-line)
+      - [Configuring the Jobs feature via the JSON configuration file](#configuring-the-jobs-feature-via-the-json-configuration-file)
+
+[*Back To The Main Readme*](../../README.md)
+
+## Jobs Feature
+
+The Jobs feature within the AWS IoT Device Client provides device-side functionality for execution of jobs created via
+https://docs.aws.amazon.com/iot/latest/developerguide/iot-jobs.html. When the Jobs feature starts within the AWS IoT Device Client,
```
KareemAliAWS, line 19: nitpick: Same as above, let's replace this hyperlink instead of displaying the full link.

source/jobs/README.md
```diff
@@ -0,0 +1,232 @@
...
+the feature will attempt to establish subscriptions to important notification topics, and will also publish
+a request to receive the latest pending jobs. It will use the shared MQTT connection to subscribe/publish to these
+topics. 
+
+When a new Job is received via these topics, the Jobs feature will look for and extract the following required and optional elements:
+
+
+### Creating a Job
+
+[View a sample job document here](sample-job-docs/install-packages.json)
```
KareemAliAWS, line 29: I get 404 and I try this link

```
https://github.com/awslabs/aws-iot-device-client/blob/206759a6264a23a52541867580d0953bc41d8541/source/jobs/sample-job-docs/install-packages.json
```

source/jobs/README.md
```diff
@@ -0,0 +1,232 @@
...
+If we wanted our remote device to print out "Hello World", then our job document so far might look like:
+```
+{
+    "operation": "echo",
+    "args": ["Hello world!"]
+}
+```
+
+`path` *string* (Optional): This field tells the Jobs feature where it should look to find an executable that matches the specified
+operation. If this field is omitted, the Jobs feature will assume that the executable can be found in its path. If
```
KareemAliAWS, line 54: Do you mean the PATH environment variable? or some custom path in the device client?

source/jobs/README.md
```diff
@@ -0,0 +1,232 @@
...
+...
+```
+The path could also be specified as a particular directory that is expected on the device, such as:
+```
+...
+"path": "/home/ubuntu/my-job-handler-folder"
+...
+```
+
+`includeStdOut` *boolean* (Optional): This field tells the Jobs feature whether the standard output (STDOUT) of the child process
```
KareemAliAWS, line 70: Not related to the README but I'm curious, since there is a size limit on the MQTT message, what happens when the stdout is too big and can't fit in a single MQTT message?

source/jobs/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+For example, we might want to create a new application written in Python called "foo" that should be run whenever our
+device receives a "foo" job. After writing the foo application, which for this particular example might only consist of a single
+file called "Foo.py", here's the steps that we would need to go through to set our AWS IoT Device Client to automatically trigger
+foo:
+
+1. Add Foo.py to our Job Handler Directory: By default, the Jobs feature will look at ~/.aws-iot-device-client/jobs/ for a handler, 
+so we should put Foo.py here (where `~/` represents the home directory of the user that will run the Device Client - /root if we're
+running the AWS IoT Device Client as a service, or /home/ubuntu/ if we're running under the default Ubuntu user). The new path
+then looks something like /home/ubuntu/.aws-iot-device-client/jobs/Foo.py.
```
KareemAliAWS, line 103: nitpick: Let's put the paths, commands and file names between tildes so that they're formatted appropriately.

source/jobs/README.md
```diff
@@ -0,0 +1,232 @@
...
+that any lines of output on STDERR will cause the job to be marked as failed. For example: 
+```
+...
+"allowStdErr": 2,
+...
+```
+This job document indicates that the Jobs feature will mark the job as a success if the return code of the process is 0 AND
+there are fewer than or equal to 2 (two) lines of standard error output from the job handler. 
+
+### Creating your own Job Handler
```
KareemAliAWS, line 89: Should we also mention that we provide some handlers by default and list/link them?

source/logging/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+The AWS IoT Device Client also provides the ability to enable native SDK logging from the AWS Common Runtime (CRT). By
+default, enabling this functionality will log TRACE level logs and above to `/var/log/aws-iot-device-client/sdk.log`
+but these defaults can be overridden as show below. If no configuration options are passed, SDK logging is disabled
+by default.
+
+### Logging Configuration Options
+
+#### Configuring the logger via the command line
+```
```
KareemAliAWS, line 30: In addition to the examples, I think we should add small/quick descriptions for each option. Some are pretty clear but it won't hurt to add quick description and list of allowed values for each of them, but some are a little bit vague like `log-type` for example begs the question what are the other possible types? how to use them ..etc.

cmake-toolchain/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+[*Back To The Main Readme*](../README.md)
+
+## Building from one architecture to the other
+
+**Description**:
+Cross Compiling allows you to compile for a **specific architecture** while on a **different architecture**.  This is extremely useful, as some *(Especially really constrained)* devices won't have the resources to complete the build process.  Even if your devices can complete the build, it is often times much easier to build in a central location (like your developer laptop) and then distribute the artifacts to appropriate devices. 
+
+The `cmake-toolchain` folder contains toolchains that will make cross compiling for other architectures easier. Currently we have toolchains to support cross compiling for the following architectures; **MIPS32**, **ARMhf**, & **AArch64**.
```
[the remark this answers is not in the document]
MiaoZhangAWS: **MIPS32**, **ARMhf**, & **AArch64**.  I think we should be more professional here with wording and avoid using '&'. Suggest to change to: 
**MIPS32**, **ARMhf**,  and **AArch64**

cmake-toolchain/README.md
```diff
@@ -0,0 +1,40 @@
+# Cross Compiliation
+ **Notice:** Running the AWS IoT Device Client will incur usage of AWS IoT services, and is likely to incur charges on your AWS account. Please refer the pricing pages for [AWS IoT Core](https://aws.amazon.com/iot-core/pricing/), [AWS IoT Device Management](https://aws.amazon.com/iot-device-management/pricing/), and [AWS IoT Device Defender](https://aws.amazon.com/iot-device-defender/pricing/) for more details.
+
+[*Back To The Main Readme*](../README.md)
+
+## Building from one architecture to the other
+
+**Description**:
+Cross Compiling allows you to compile for a **specific architecture** while on a **different architecture**.  This is extremely useful, as some *(Especially really constrained)* devices won't have the resources to complete the build process.  Even if your devices can complete the build, it is often times much easier to build in a central location (like your developer laptop) and then distribute the artifacts to appropriate devices.
```
MiaoZhangAWS, line 9: Cross Compiling -> Cross-compiling. 
This sentence confused me more about what cross compilation is. I think it needs better wording and introduction here. If someone is new to this concept, it will be hard for him to follow. You can also link some good definition you find from other place.

cmake-toolchain/README.md
```diff
@@ -0,0 +1,40 @@
...
+The `cmake-toolchain` folder contains toolchains that will make cross compiling for other architectures easier. Currently we have toolchains to support cross compiling for the following architectures; **MIPS32**, **ARMhf**, & **AArch64**.
+
+You can specify one of the given toolchains when running cmake:
+```
+cmake ../ -DCMAKE_TOOLCHAIN_FILE=<Path/To/Toolchain>
+```
+This will allow the toolchain to overwrite variables *(various paths, compilers, & flags)* required to execute cross compilation without changing the original **cmake** file.
+
+**Dependencies**:
+For your build to be successful you'll also need a cross compiled version of our dependencies ([aws-iot-device-sdk-cpp-v2](https://github.com/aws/aws-iot-device-sdk-cpp-v2) & **openssl**), for the SDK this is automatically accomplished when running the **cmake** command above without the following flag `-DBUILD_SDK=OFF`.
```
MiaoZhangAWS, line 20: Same with above. Change **_&_** to **_and_**.

MiaoZhangAWS (commented):
I think this documentation update can benefit from having a doc writer review the wording and polishing the format.

cmake-toolchain/README.md
MiaoZhangAWS: Can you list the full command here? Instead just giving `-DBUILD_SDK=OFF`

cmake-toolchain/README.md
```diff
@@ -0,0 +1,40 @@
+# Cross Compiliation
```
MiaoZhangAWS, line 1: Is there any trouble shooting advice or common errors you can provide? It will help user with debugging.

source/tunneling/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+On the remote device:
+1. Deploy and run the Device Client with Secure Tunneling feature enabled
+
+On your laptop:
+1. Create a new secure tunnel on the AWS Console
+2. Copy the source access token
+3. Start the local proxy with the source access token. For example:  
+```
+$ ./localproxy -r us-east-1 -s 8080 -t <source token>
```
MiaoZhangAWS, line 51: Is device client going to support starting local proxy in source mode? 
Remember from our last discussion, it's more for running local proxy in destination mode on remote device.

MiaoZhangAWS: We should post [this link](https://github.com/aws-samples/aws-iot-securetunneling-localproxy/blob/master/README.md#options-set-via-command-line-arguments) to show what other inputs can be used to start local proxy.

README.md
```diff
@@ -102,7 +67,7 @@ The AWS IoT Device Client currently works by default on IoT devices with common
 To use the AWS IoT Device Client, you'll need to compile an executable using the source code provided by this
```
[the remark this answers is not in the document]
lwilkovich: You could compile w/ them static or shared, that's up to the customer.  Regardless of location required during runtime

cmake-toolchain/README.md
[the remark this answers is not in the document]
lwilkovich: The only thing they would be required to download is whatever specific compiler is required

README.md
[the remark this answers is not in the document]
KareemAliAWS: I see, so they are a requirement, in that case I think that part `Note: CI Build uses` is confusing(at least for me), would probably be clearer to just write

### Minimum Requirements

* C++ 11 or higher
* [CMake](https://cmake.org/) 3.10+
* OpenSSL 1.1.1
* [aws-iot-device-sdk-cpp-v2](https://github.com/aws/aws-iot-device-sdk-cpp-v2) commit hash located in `CMakeLists.txt.awssdk`

cmake-toolchain/README.md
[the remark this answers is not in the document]
KareemAliAWS: In that case, I don't think that qualifies as providing the toolchain since the toolchain is basically the compiler, assembler, debugger, linker, ...etc (other tools you need to produce the executable), the things that we don't provide and require the users to download themselves.

source/jobs/README.md
lwilkovich: this is a variable in the job docs file(s)

source/logging/README.md
lwilkovich: The options are mentioned in the first paragraph

README.md
```diff
@@ -3,69 +3,27 @@
...
 ## Introduction
+*__Sections:__*
+* [Current Capabilities](#current-capabilities) 
+* [List of Supported Platforms](#list-of-supported-platforms)
+
+[*Back To The Top*](#aws-iot-device-client)
+
 
 The AWS IoT Device Client is free, open-source, modular software written in C++ that you can compile and install on
  your Embedded Linux based IoT devices to access [AWS IoT Core](https://aws.amazon.com/iot-core/), [AWS IoT Device
```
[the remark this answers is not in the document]
lwilkovich: I'll look into this separately

README.md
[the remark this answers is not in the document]
lwilkovich: So it's complicated because you could potentially use something else besides Openssl entirely if you so choose.  Similarly there is a potential that different versions of the sdk could  work fine

cmake-toolchain/README.md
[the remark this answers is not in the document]
lwilkovich: I changed it to call them tool chain files, and added a note that the compiler is not included in this

docs/PERMISSIONS.md
```diff
@@ -0,0 +1,39 @@
...
+
+#### Recommended and Required permissions on directories storing respective files
+Directory     | Chmod Permissions | Required |
+------------- | ------------- | ------------- 
+Directory Storing Private Key | 700 | **Yes**
+Directory Storing Public Certificates   | 700 | **Yes**
+Directory Storing Root Certificate Authority | 700 | **Yes**
+Directory Storing CSR File  | 700 | **Yes**
+Directory Storing Log File  | 745 | **Yes**
+Directory Storing Config Files | 745 | **Recommended**
```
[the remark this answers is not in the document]
KareemAliAWS: If they're not required, what are we warning the users about?

source/jobs/README.md
KareemAliAWS: Sorry, should've clarified I was referring to the `path` in the sentence, `If this field is omitted, the Jobs feature will assume that the executable can be found in its path ` not the `path` in `path *string* (Optional): This field tells the Jobs feature where it should`

source/logging/README.md
KareemAliAWS: Ah, I missed that, it's probably better to use the format used in other README files, for example [here](https://github.com/awslabs/aws-iot-device-client/tree/206759a6264a23a52541867580d0953bc41d8541/source/devicedefender#device-defender-feature-configuration-options) but I guess it's fine since it's in the description.

README.md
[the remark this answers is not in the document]
KareemAliAWS: Cool, Thanks!

source/jobs/README.md
lwilkovich: Oh yes ill update that, in the sentence that is the environment variable PATH

README.md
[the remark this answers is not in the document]
KareemAliAWS: I think (and correct me if I'm wrong) that these requirements are about what we support not what could be done, for example I'm pretty sure there are tools that can convert c++11 code to c++98 compatible code but we still say clearly that c++11 or hight is required, so if we support multiple TLS implementations, that's fine, and we can specify which ones we support building with any commit we can say that, if we support only subset of commits, then let's specify that as well. But we shouldn't leave it vague like that.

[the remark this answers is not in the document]
lwilkovich: We don't currently have infrastructure for testing other TLS stacks or versions of the SDK so we don't really know for certain.  Which is why I think it's better to be vague, more of a "If you want to try it, it might work but we don't actively support that in testing or our CI"  

Also for 11->98 I've never heard of anything like that, at least nothing that works well.  Our requirement on c++11 was chosen because the SDK is the same.  Cmake aswell,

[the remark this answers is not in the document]
KareemAliAWS: Then we need to ask ourselves and figure out what implementation do we support, leaving it vague is just plain confusing.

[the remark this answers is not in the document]
lwilkovich: I suppose we could change it to match your earlier message.  and then a note at the end that other TLS stacks and versions could work but we don't actively support them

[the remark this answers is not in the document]
KareemAliAWS: Yeah, I think that would make it clear. Thanks!

KareemAliAWS (approved):

xps506 (approved):

PR #125, opened 2021-03-23
Fix mac build failure caused by linking s2n

matthewpaul:
According to the SDK team, we should not be linking against s2n when we're building on Mac/OS X. This commit modified the build instructions to avoid linking against s2n if the target platform is Mac.


KareemAliAWS (dismissed):

KareemAliAWS (commented):
Should we have an action item to add a CI check for macos as well?

lwilkovich (dismissed):

matthewpaul:
@KareemAliAWS Yes I have filed a ticket to address that

DavidOgunsAWS (approved):

MiaoZhangAWS (approved):

Issue #103, opened 2021-02-01
Add support to pass 'Device Configuration' data down to the device when using AWS IoT Fleet Provisioning


janborch:
The AWS IoT Fleet provisioning feature allows arbitrary data to be passed down to the device after the cloud-side provisioning as explained here: https://docs.aws.amazon.com/iot/latest/developerguide/provision-template.html#device-config

AWS IoT Device client currently does not make this data accessible to the device. 

An option is to append the 'Device Configuration'  data to the `runtime-config` file to allow other software running on the device to read the data.

```
{
"runtime-config": {
    "completed-fp": true,
    "cert": "/root/.aws-iot-device-client/keys/xxxxxxxxxxxxxxx-certificate.pem.crt",
    "key": "/home/ubuntu/device.key",
    "thing-name": "ThingName01",
    "device-configuration":{
             "Foo": "Bar",
             "customer-id": "1234556"
    }
    }
}

```

HarshGandhi-AWS:
Hello @janborch ,
 
Thank you for suggesting us this enhancement in Fleet Provisioning feature. As requested we have [made changes](https://github.com/awslabs/aws-iot-device-client/pull/141) in the code-base for providing support for storing Fleet Provisioning configurations passed via Fleet Provisioning template in the runtime-config file. 
I am closing this issue now, please feel free to reopen this issue or create a new one if required. 

Thank you,
Harsh Gandhi

PR #141, opened 2021-04-23
Added support for storing FP device config from FP Template in runtime-config file

HarshGandhi-AWS:
### Motivation
- Important feature for all device on-boarding flows where the device will store device config provided via FP template in runtime-config
- Issue number: #103 
- More Details: 
 The AWS IoT Fleet provisioning feature allows arbitrary data to be passed down to the device after the cloud-side provisioning as explained here: https://docs.aws.amazon.com/iot/latest/developerguide/provision-template.html#device-config

#### Change summary
With this change I am appending the 'Device Configuration'  data to the `runtime-config` file to allow other software running on the device to read the data.

```
{
"runtime-config": {
    "completed-fp": true,
    "cert": "/root/.aws-iot-device-client/keys/xxxxxxxxxxxxxxx-certificate.pem.crt",
    "key": "/home/ubuntu/device.key",
    "thing-name": "ThingName01",
    "device-configuration":{
             "Foo": "Bar",
             "customer-id": "1234556"
    }
    }
}

``` 

#### Revision diff summary

N/A

### Testing
- Yes, this change is tested. I have manually made changes in FP template and provisioned the device to validate changes. 
- Also I have added required unit tests for new string util method.


source/fleetprovisioning/FleetProvisioning.h
```diff
@@ -115,10 +115,17 @@ namespace Aws
                      * \brief the location where keys generated by Fleet Provisioning will be stored
                      */
                     std::string keyDir = Config::DEFAULT_KEY_DIR;
+
+                    /**
+                     * \brief stores device config of newly provisioned device
```
matthewpaul, line 120: What is a device config?

source/util/StringUtils.cpp
```diff
@@ -60,6 +61,31 @@ namespace Aws
...
+
+                string MapToJsonString(Crt::Optional<Crt::Map<Aws::Crt::String, Aws::Crt::String>> map)
+                {
+                    std::map<Aws::Crt::String, Aws::Crt::String>::iterator it;
+                    string result = "";
+                    unsigned int count = 0;
+                    for (it = map->begin(); it != map->end(); it++)
+                    {
+                        count++;
+                        result.erase(std::find(result.begin(), result.end(), '\0'), result.end());
```
matthewpaul, line 79: why do you need to run this erase on each iteration of the loop?

source/util/StringUtils.cpp
```diff
@@ -60,6 +61,31 @@ namespace Aws
...
+                        result = result.append(addString(it->first, it->second));
+                        if (count != map->size())
+                        {
+                            result.erase(std::find(result.begin(), result.end(), '\0'), result.end());
```
matthewpaul, line 83: and again here

source/util/StringUtils.h
```diff
@@ -39,6 +40,23 @@ namespace Aws
...
+                 * \brief helper method to concat strings with ':' in between
+                 *
+                 * @param first first string
+                 * @param second second string
+                 * @return string in 'first:second' form
+                 */
+
+                std::string addString(Aws::Crt::String first, Aws::Crt::String second);
+                /**
+                 * \brief Given an input map object, will return the string form of the map
```
matthewpaul, line 54: is it in JSON format? Might be helpful to state whether this is a valid JSON object or not

source/util/StringUtils.cpp
HarshGandhi-AWS: I need to remove the null terminator character '\0'. Without doing that the next lines or string will not be written in the file. 
Example: For 
{a:b,
c:d}. Only a:b was written in the file.

source/util/StringUtils.h
HarshGandhi-AWS: The input coming from FP template is validated to be JSON by the API. So before the response is available for us, Device Config is validated to be in JSON format and then converted into Map object.

source/fleetprovisioning/FleetProvisioning.h
HarshGandhi-AWS: The device configuration contains arbitrary data user want to send to your devices when provisioning. More details are available [here](https://docs.aws.amazon.com/iot/latest/developerguide/provision-template.html#device-config).

source/util/StringUtils.cpp
lwilkovich: Can \0 exist in a key-pair

lwilkovich: Also shouldn't \0 just be removed at the end of the entire operation?

source/fleetprovisioning/FleetProvisioning.h
matthewpaul: Maybe include some of that info in this doxygen comment? Otherwise the reader cannot get any valuable info from this comment

source/util/StringUtils.cpp
matthewpaul: It seems like maybe it should just be run here?

source/util/StringUtils.h
matthewpaul: Ok, but is your function generating a valid JSON object? That might be important information for someone reading the API documentation here

HarshGandhi-AWS: No, it will return String form of the map object. I thing the method name can confuse users, will update that to 'MapToString'

A good example can is this [maptoString](https://github.com/awslabs/aws-iot-device-client/pull/141/files#diff-e68e49b78fd339a42c1c8b65419a470b5d3af97731474c4ba5c3a02ac0f494fe) Unit test I have added in TestStringUtils.cpp file.

source/fleetprovisioning/FleetProvisioning.h
HarshGandhi-AWS: Done.

source/util/StringUtils.cpp
HarshGandhi-AWS: Updated code to remove all '\0' in the end.

matthewpaul (approved):

PR #142, opened 2021-04-26
Updated LoadFromJson function for pubsub sample to resolve seg fault

HarshGandhi-AWS:
### Motivation
- With Pub-Sub sample disabled and config items with empty value in the pub-sub block, the DC application was failing to start with and segmentation fault. 
- ExportDefaultSetting method was not updated with our default config file provided to our customers. 

```
./aws-iot-device-client 
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted (core dumped)
```

#### Change summary
- Updated LoadFromJson function for pub-sub sample to resolve segmentation fault 
- Updated ExportDefaultSetting to sync with our default 'config-template.json' file. Added configs for Logger and FP.

#### Revision diff summary

- Removed Error Logs from validation method to avoid confusion as asked.

### Testing
 This change is tested and verified manually.

### DIFF
--- a/source/config/Config.cpp
+++ b/source/config/Config.cpp
@@ -1044,10 +1044,6 @@ bool PlainConfig::PubSub::Validate() const
     {
         if (!FileUtils::ValidateFilePermissions(publishFile.value(), Permissions::PUB_SUB_FILES, true))
         {
-            LOGM_ERROR(
-                Config::TAG,
-                "*** %s: Publish File field must be specified if Pub-Sub sample feature is enabled ***",
-                DeviceClient::DC_FATAL_ERROR);
             return false;
         }
     }
@@ -1063,10 +1059,6 @@ bool PlainConfig::PubSub::Validate() const
     {
         if (!FileUtils::ValidateFilePermissions(subscribeFile.value(), Permissions::PUB_SUB_FILES, true))
         {
-            LOGM_ERROR(
-                Config::TAG,
-                "*** %s: Subscribe File field must be specified if Pub-Sub sample feature is enabled ***",
-                DeviceClient::DC_FATAL_ERROR);
             return false;
         }
     }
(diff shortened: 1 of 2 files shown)


source/config/Config.cpp
```diff
@@ -1012,6 +1040,17 @@ bool PlainConfig::PubSub::Validate() const
             DeviceClient::DC_FATAL_ERROR);
         return false;
     }
+    if (publishFile.has_value() && !publishFile->empty())
+    {
+        if (!FileUtils::ValidateFilePermissions(publishFile.value(), Permissions::PUB_SUB_FILES, true))
+        {
+            LOGM_ERROR(
+                Config::TAG,
+                "*** %s: Publish File field must be specified if Pub-Sub sample feature is enabled ***",
```
lwilkovich, line 1049: This might be confusing because if we made it here, something was specified... it just failed the file permission check

source/config/Config.cpp
```diff
@@ -1020,17 +1059,14 @@ bool PlainConfig::PubSub::Validate() const
...
-            return false;
-        }
-    }
     if (subscribeFile.has_value() && !subscribeFile->empty())
     {
         if (!FileUtils::ValidateFilePermissions(subscribeFile.value(), Permissions::PUB_SUB_FILES, true))
         {
+            LOGM_ERROR(
+                Config::TAG,
+                "*** %s: Subscribe File field must be specified if Pub-Sub sample feature is enabled ***",
```
matthewpaul, line 1068: Why is this being enforced? In the README we state that the pub sub feature does not need to have files specified, and will just publish Hello World and log a message

lwilkovich: It only enters this if something is specified & it's "bad"

source/config/Config.cpp
HarshGandhi-AWS: Noted. Will remove it.

matthewpaul (approved):

lwilkovich (approved):

PR #149, opened 2021-06-16
fix the high cpu utilization issue caused by infinite while loop in StdoutLogger

xps506:
### Motivation
- We found the issue that it will take almost 99% of CPU when device client is running, this change is going to fix this by adding sleep time in an infinite while loop we use in StdoutLogger. 
- Issue number:  https://github.com/awslabs/aws-iot-device-client/issues/14

#### Change summary
This a one-line change to make logger thread sleep 25ms after every time we fetch a log message from log queue

#### Revision diff summary

### Testing
 **Is your change tested? If not, please justify the reason.**  I've tested this change on my personal testing machines and I verified that CPU utilization would be down to 1% after I add sleep time

### DIFF
--- a/source/logging/StdOutLogger.cpp
+++ b/source/logging/StdOutLogger.cpp
@@ -31,6 +31,7 @@ void StdOutLogger::run()
         {
             writeLogMessage(std::move(message));
         }
+        this_thread::sleep_for(std::chrono::milliseconds(25));
     }
 }
 
--- a/source/logging/StdOutLogger.cpp
+++ b/source/logging/StdOutLogger.cpp
@@ -12,6 +12,8 @@
 using namespace std;
 using namespace Aws::Iot::DeviceClient::Logging;
 
+constexpr int StdOutLogger::DEFAULT_WAIT_TIME_MILLISECONDS;
+
 void StdOutLogger::writeLogMessage(unique_ptr<LogMessage> message)
 {
     char time_buffer[TIMESTAMP_BUFFER_SIZE];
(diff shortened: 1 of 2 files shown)


HarshGandhi-AWS (approved):
LGTM

xps506:
> https://issues.amazon.com/issues/DC-1965 is an internal link, other Github users won't able to open it. Suggest to remove it to avoid confusion.

updated

source/logging/StdOutLogger.cpp
```diff
@@ -31,6 +31,7 @@ void StdOutLogger::run()
         {
             writeLogMessage(std::move(message));
         }
+        this_thread::sleep_for(std::chrono::milliseconds(25));
```
MiaoZhangAWS, line 34: How is the 25 determined? 
Have we verified the sleep time cause any impact on the logging ability of device client? For example, the logs writing has delay?

kareali, line 34: Is there a reason for choosing 25 millisecond specifically? Also Could you use a preprocessor directive constant to define this value?

xps506: I've tested this, logging functionality doesn't get impacted by this sleep time. If we set sleep time too low the cpu utilization will increase but overall I've tested 25ms works fine  my testing machine

xps506: I couldn't find a best value this but I'm now trying to lower the sleep time to 1ms and it also works well but get a slight increase cpu utilization from 0.3% to 1%, basically the shorter sleep time the high cpu utilization but overall it seems all value works as long as we add sleep time and I have verified the logging functionality doesn't get impacted by this change

xps506: I update to set constant define value for this sleep time

MiaoZhangAWS (approved):

HarshGandhi-AWS (approved):
LGTM

Issue #147, opened 2021-06-08
Device client using greater than 99% CPU usage.


dhwalters423:
**Describe the bug**
Device client is using greater than 98% CPU usage consistently on the NXP S32G with Yocto Linux

**To Reproduce**
```
./aws-iot-device-client --config-file /etc/aws-iot-device-client.json
```

**Expected behavior**

Device should use less than 10% CPU usage

**Actual behavior**

Device Client is using consistently between 98-99% CPU usage

**Logs**

```
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
ps aux | grep device
root         338 99.9  0.1 306316  7320 pts/0    Sl   14:54  15:05 ./aws-iot-device-client --config-file /etc/aws-iot-device-client.json
root         358  0.0  0.0   3188   820 pts/0    S+   15:09   0:00 grep device
```

```
2021-06-08T15:25:04.794Z [INFO]  {Main.cpp}: Now running AWS IoT Device Client version v1.1.0
2021-06-08T15:25:04.799Z [INFO]  {SharedCrtResourceManager.cpp}: SDK logging is disabled. Enable it with --enable-sdk-logging on the command line or logging::enable-sdk-logging in your configuration file
2021-06-08T15:25:04.800Z [DEBUG] {Retry.cpp}: Retryable function will retry until success
2021-06-08T15:25:04.818Z [INFO]  {SharedCrtResourceManager.cpp}: Establishing MQTT connection with client id S32_Secure_Tunneling_Thing...
2021-06-08T15:25:05.439Z [INFO]  {SharedCrtResourceManager.cpp}: MQTT connection established with return code: 0
2021-06-08T15:25:05.439Z [INFO]  {SharedCrtResourceManager.cpp}: Shared MQTT connection is ready!
2021-06-08T15:25:05.440Z [INFO]  {Main.cpp}: Jobs is disabled
2021-06-08T15:25:05.440Z [INFO]  {Main.cpp}: Secure Tunneling is disabled
2021-06-08T15:25:05.440Z [INFO]  {Main.cpp}: Device Defender is disabled
2021-06-08T15:25:05.440Z [INFO]  {Main.cpp}: Pub Sub is disabled
2021-06-08T15:25:06.586Z [INFO]  {Main.cpp}: Received signal: (2)
[...]

PR #150, opened 2021-06-17
Another fix of the high cpu utilization issue caused by infinite while loop in FileLogger

xps506:
### Motivation
- We found the issue that it will take almost 99% of CPU when device client is running, I had made a change on StdLogger file but this change is going to fix this by adding sleep time in an infinite while loop we use in FileLogger
- Issue number: https://github.com/awslabs/aws-iot-device-client/issues/147

#### Change summary
This a one-line change to make logger thread sleep 1ms after every time we fetch a log message from log queue

#### Revision diff summary

### Testing
 **Is your change tested? If not, please justify the reason. I've tested this change on my personal testing machines by running DC as service with FileLogger enabled and I verified that CPU utilization would be down to 1% after I add sleep time
- CI test run result: <link>

### DIFF
--- a/source/logging/FileLogger.cpp
+++ b/source/logging/FileLogger.cpp
@@ -14,6 +14,8 @@ using namespace std;
 using namespace Aws::Iot::DeviceClient::Logging;
 using namespace Aws::Iot::DeviceClient::Util;
 
+constexpr int FileLogger::DEFAULT_WAIT_TIME_MILLISECONDS;
+
 bool FileLogger::start(const PlainConfig &config)
 {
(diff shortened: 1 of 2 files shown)


HarshGandhi-AWS (approved):

MiaoZhangAWS (commented):
Notice the functions in _FileLogger.cpp_ and _StdOutLogger.cpp_ has lots of duplication. 
It will benefit from minor refactoring to extract the common functions into a base class/interface. And let FileLogger and StdOutLogger inherit from this base class/interface. 

Not a blocker to approve this change.

MiaoZhangAWS (approved):

PR #166, opened 2021-08-24
Added job handler for downloading specified files

HarshGandhi-AWS:
### Motivation
- Add job handler and sample job document for downloading user specified files.
- This job handler will also support downloading of S3 document using presign URLs.

#### Change summary

#### Revision diff summary

### Testing
- Manually tested job handler locally.

### DIFF
--- a/sample-job-docs/download-file.json
+++ b/sample-job-docs/download-file.json
@@ -1,0 +1,6 @@
+{
+  "_comment": "This sample JSON file can be used for downloading specific file from cloud storage. The arguments passed here in `args` are valid file URL and path where we want to store the downloaded file on the device.",
+  "operation": "download-file.sh",
+  "args": ["https://github.com/awslabs/aws-iot-device-client/archive/refs/tags/v1.3.tar.gz", "/tmp/aws-iot-device-client.tar.gz"],
+  "path": "default"
+}
--- a/sample-job-handlers/download-file.sh
+++ b/sample-job-handlers/download-file.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+set -e
+
+echo "Running download-file.sh"
+if command -v "wget" > /dev/null
+then
+    echo "Using wget for downloading user specified file"
+    if command -v "sudo" > /dev/null; then
+      sudo -n wget --quiet -O "$2" "$1"
+    else
+      echo "WARNING: sudo command not found"
+      wget --quiet -O --quiet "$2" "$1"
+    fi
+else
+    >&2 echo "wget software not installed on the device. Use install-packages operation for installing wget software on the target device."
+    exit 1
+fi


sample-job-docs/download-file.json
```diff
@@ -0,0 +1,6 @@
+{
+  "_comment": "This sample JSON file can be used for downloading specific file from cloud storage. The arguments passed here in `args` are valid file URL and path where we want to store the downloaded file on the device.",
+  "operation": "download-file.sh",
+  "args": ["https://github.com/awslabs/aws-iot-device-client/archive/refs/tags/v1.3.tar.gz", "/tmp/aws-iot-device-client.tar.gz"],
```
satheshantanu, line 4: To avoid customer confusion I'd name this file something else. 
Perhaps something like "tmp/Downloaded_File.tar.gz"?

sample-job-handlers/download-file.sh
```diff
@@ -0,0 +1,17 @@
...
+then
+    echo "Using wget for downloading user specified file"
+    if command -v "sudo" > /dev/null; then
+      sudo -n wget --quiet -O "$2" "$1"
+    else
+      echo "WARNING: sudo command not found"
+      wget --quiet -O --quiet "$2" "$1"
+    fi
+else
+    >&2 echo "wget software not installed on the device. Use install-packages operation for installing wget software on the target device."
```
satheshantanu, line 15: "Wget software package not installed on the device. Use the "install-packages.sh" operation to install the wget software package on this device."

sample-job-handlers/download-file.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+set -e
+
+echo "Running download-file.sh"
+if command -v "wget" > /dev/null
```
satheshantanu, line 5: Are there any other mechanisms that customers use? CURL?

satheshantanu: Do we want to make this optional (i.e., use either curl or wget?)

HarshGandhi-AWS: There is an issue with using CURL with S3 pre-signed URL's. If the user does not have the access to download the S3 file, instead of throwing an error and aborting operation with non-zero exit code, CURL will return exit code `0` (success) resulting in DC updating the Jobs status to `SUCCESS`. For some reason, CURL will write the exception in the output-document.  Because of this issue, user will not be notified that the download-file.sh operation failed. 
On the other hand, if wget sees access denied exception, it will abort the program with `6` and the Job status will be updated to `FAIL`.

sample-job-docs/download-file.json
HarshGandhi-AWS: Done.

sample-job-handlers/download-file.sh
HarshGandhi-AWS: Done.

sample-job-handlers/download-file.sh
satheshantanu: Are there no other indications we can use to identify whether the download using CURL was successful? Like perhaps compare the filesize of the URL loc to the output file?

bisasmit (approved):

sample-job-handlers/download-file.sh
HarshGandhi-AWS: To validate file data, we need to parse the file and check the contents. But that would get complicated because the file type can be different.
To validate based on size, we need to ask user to give us original file size data. So that option is also not the best solution.

satheshantanu (approved):

PR #170, opened 2021-09-10
Adding Validation for config file and cli inputs

xlcheng1:
### Motivation
- Issue number: 

#### Change summary

#### Revision diff summary

### Testing
- CI test run result: <link>


source/util/FileUtils.cpp
```diff
@@ -302,3 +302,26 @@ bool FileUtils::FileExists(const string &filename)
     ifstream f(expandedPath);
     return f.good();
 }
+
+bool FileUtils::IsValidFilePath(const string &filePath)
+{
+    wordexp_t word;
+    switch (wordexp(filePath.c_str(), &word, 0))
```
HarshGandhi-AWS, line 309: I think at this line of code we will have segmentation fault if the file path is incorrect.

source/util/FileUtils.cpp
```diff
@@ -302,3 +302,26 @@ bool FileUtils::FileExists(const string &filename)
     ifstream f(expandedPath);
     return f.good();
 }
```
HarshGandhi-AWS, line 304: Add another function to validate if the file path does not have any illegal filePath characters.

source/util/FileUtils.cpp
```diff
@@ -302,3 +302,25 @@ bool FileUtils::FileExists(const string &filename)
...
+
+bool FileUtils::IsValidFilePath(const string &filePath)
+{
+    wordexp_t word;
+    switch (wordexp(filePath.c_str(), &word, 0))
+    {
+        case 0:
+            break;
+        case WRDE_BADCHAR:
```
HarshGandhi-AWS, line 313: replace this with `WRDE_NOSPACE` and have a default case to return false so all other failure cases will be covered. 

```
case 0: 
    break;
case WRDE_NOSPACE:
      wordfree (&word);
default:                    /* Some other error.  */
      return false;
```

test/config/TestConfig.cpp
```diff
@@ -14,14 +14,22 @@ using namespace Aws::Crt;
...
+
 TEST(Config, AllFeaturesEnabled)
 {
     constexpr char jsonString[] = R"(
 {
     "endpoint": "endpoint value",
-    "cert": "cert",
-    "key": "key",
-    "root-ca": "root-ca",
+    "cert": "/tmp/test",
```
HarshGandhi-AWS, line 30: update this to "/tmp/aws-iot-device-client-test-file" since it will be created in users linux machine, it is better to name it after device client.

test/config/TestConfig.cpp
```diff
@@ -14,14 +14,22 @@ using namespace Aws::Crt;
 using namespace Aws::Iot::DeviceClient;
 using namespace Aws::Iot::DeviceClient::Util;
 
+TEST(Config, CreateTempFile)
+{
+    string filePath = "/tmp/test";
```
HarshGandhi-AWS, line 19: Consider saving the file path in some variable if possible since it will be used everywhere in this file and also that way it would be easier to update the file path.

test/config/TestConfig.cpp
```diff
@@ -70,9 +78,9 @@ TEST(Config, AllFeaturesEnabled)
 
     ASSERT_TRUE(config.Validate());
     ASSERT_STREQ("endpoint value", config.endpoint->c_str());
-    ASSERT_STREQ("cert", config.cert->c_str());
-    ASSERT_STREQ("key", config.key->c_str());
-    ASSERT_STREQ("root-ca", config.rootCa->c_str());
+    ASSERT_STREQ("/tmp/test", config.cert->c_str());
```
HarshGandhi-AWS, line 81: Consider saving the file path in some variable if possible since it will be used everywhere in this file and also that way it would be easier to update the file path.

source/util/FileUtils.cpp
```diff
@@ -302,3 +302,27 @@ bool FileUtils::FileExists(const string &filename)
     ifstream f(expandedPath);
     return f.good();
 }
+
+bool FileUtils::IsValidFilePath(const string &filePath)
+{
```
HarshGandhi-AWS, line 307: I would suggest adding comments in this method so you do not have to write comments in code wherever we call `IsValidFilePath` method.

source/config/Config.cpp
```diff
@@ -1278,23 +1335,45 @@ bool PlainConfig::SampleShadow::Validate() const
...
     }
 
     if (shadowOutputFile.has_value() && !shadowOutputFile->empty())
     {
-        if (!FileUtils::ValidateFilePermissions(shadowOutputFile.value(), Permissions::SAMPLE_SHADOW_FILES, true))
+        if (FileUtils::IsValidFilePath(shadowOutputFile->c_str()))
+        {
+            if (!FileUtils::ValidateFilePermissions(shadowOutputFile.value(), Permissions::SAMPLE_SHADOW_FILES, true))
+            {
+                LOGM_ERROR(
```
HarshGandhi-AWS, line 1365: Move these comments in `IsValidFilePath` method.

source/config/Config.cpp
```diff
@@ -622,6 +634,14 @@ bool PlainConfig::Jobs::LoadFromCliArgs(const CliArgs &cliArgs)
 
 bool PlainConfig::Jobs::Validate() const
 {
+    if (!handlerDir.empty())
```
HarshGandhi-AWS, line 637: Do we want to add has_value check over here? Otherwise we might see an segmentation fault for checking is empty on null value.

source/config/Config.cpp
```diff
@@ -1133,12 +1176,19 @@ bool PlainConfig::PubSub::Validate() const
     }
     if (subscribeFile.has_value() && !subscribeFile->empty())
     {
-        if (!FileUtils::ValidateFilePermissions(subscribeFile.value(), Permissions::PUB_SUB_FILES, true))
+        if (FileUtils::IsValidFilePath(subscribeFile->c_str()))
+        {
+            if (!FileUtils::ValidateFilePermissions(subscribeFile.value(), Permissions::PUB_SUB_FILES, true))
+            {
+                LOGM_ERROR(
```
HarshGandhi-AWS, line 1183: This log statement is redundant. We already log errors in `ValidateFilePermissions` method.

HarshGandhi-AWS (approved):
LGTM

bisasmit (approved):

PR #193, opened 2021-11-25
Handler typo

marcoemorais-aws:
### Motivation
- The `service_name` should appear before `command` in service job handlers, but are transposed.  As a result, the handler will always fail.

#### Change summary
- Reorder so that `service_name` appears before `command` in service job handlers. 

#### Revision diff summary
N/A

### Testing
- CI test run result: <link>

### DIFF
--- a/source/config/Config.cpp
+++ b/source/config/Config.cpp
@@ -615,7 +615,8 @@ bool PlainConfig::Jobs::LoadFromJson(const Crt::JsonView &json)
     }
 
     jsonKey = JSON_KEY_HANDLER_DIR;
-    if (json.ValueExists(jsonKey)) {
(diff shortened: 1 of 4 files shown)


source/config/Config.cpp
```diff
@@ -614,6 +614,12 @@ bool PlainConfig::Jobs::LoadFromJson(const Crt::JsonView &json)
         enabled = json.GetBool(jsonKey);
     }
 
+    jsonKey = JSON_KEY_HANDLER_DIR;
```
HarshGandhi-AWS, line 617: why was this added here again? I think you might not be using the latest code.

HarshGandhi-AWS (approved):

xlcheng1 (approved):

Issue #181, opened 2021-10-18
Invoking --enable-pub-sub surfaces error "Unrecognized command line argument"


rpcme:
**Describe the bug**

Cannot run pubsub with --enable-pub-sub command line argument

**To Reproduce**

```bash
./aws-iot-device-client --enable-sdk-logging                         --endpoint $(aws iot --profile work --region us-east-1 describe-endpoint --endpoint-type iot:Data-ATS --output text --query endpointAddress)                         --enable-fleet-provisioning true --enable-pub-sub true --publish-topic dc-pub-topic --subscribe-topic dc-sub-topic                       --cert ${credpath}/28eb7244-978a-4376-b81b-1bf9c0cad826.crt.pem                         --key ${credpath}/28eb7244-978a-4376-b81b-1bf9c0cad826.key.prv.pem                         --root-ca ${credpath}/root-ca.pem                         --thing-name $(uuidgen)                         --fleet-provisioning-template-name fp-template-claim 
2021-10-18T05:37:23.925Z [ERROR] {Config.cpp}: *** AWS IOT DEVICE CLIENT FATAL ERROR: Unrecognised command line argument: --enable-pub-sub ***
```

**Expected behavior**

Run the pubsub demo

**Actual behavior**

Fatal error

**Logs**

If applicable, add full logs of errors and outputs to help explain your problem. Preferably, you can also [increase the verbosity of the logs](https://github.com/awslabs/aws-iot-device-client#logging).

**Environment (please complete the following information):**
 - OS: Ubuntu
 - Version 20.04
 - Architecture: x86-64
 - Device Client commit: 0cae46162cbe2f10205bc2afbe523f368f3a7783

**Additional context**

HarshGandhi-AWS:
Hello @rpcme , again thank you for reaching out to us with this issue. @marcoemorais-aws has released the [bug fix](https://github.com/awslabs/aws-iot-device-client/pull/198) for this issue. Please review and test it to see if the issue is resolved for you or not. 
I am resolving this issue now, please feel free to reopen it if the issue is not resolved or open a new issue with us if you are facing any difficulty using the Device Client.

Regards, 
Harsh Gandhi

PR #198, opened 2021-12-08
Add PubSub features to CLI ArgumentDefinition

marcoemorais-aws:
### Motivation
- All of the pub-sub related cli arguments (pasted below) were accidentally omitted from Config::ArgumentDefinition which means that users cannot pass these arguments from the cli.
  - `--enable-pub-sub`
  - `--publish-topic`
  - `--publish-file`
  - `--subscribe-topic`
  - `--subscribe-file`
- Adds PubSub cli arguments to ArgumentDefinition.
- Issue number: https://github.com/awslabs/aws-iot-device-client/issues/181

#### Change summary
- Add pub-sub related cli arguments to Config::ArgumentDefinition.

#### Revision diff summary

### Testing
- Tested that behavior of all arguments passed using the cli is equivalent to passing in config file.

### DIFF
--- a/source/config/Config.cpp
+++ b/source/config/Config.cpp
@@ -1457,6 +1457,12 @@ bool Config::ParseCliArgs(int argc, char **argv, CliArgs &cliArgs)
         {PlainConfig::FleetProvisioning::CLI_FLEET_PROVISIONING_CSR_FILE, true, false, nullptr},
         {PlainConfig::FleetProvisioning::CLI_FLEET_PROVISIONING_DEVICE_KEY, true, false, nullptr},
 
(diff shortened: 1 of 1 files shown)


HarshGandhi-AWS (commented):
Everything looks good. I think we should add validation checks for config file values[ over here ](https://github.com/awslabs/aws-iot-device-client/blob/0cae46162cbe2f10205bc2afbe523f368f3a7783/test/config/TestConfig.cpp#L583)in this unit test. Right now we are only validating on config received from CLI.

marcoemorais-aws:
Added tests to `TestConfig` for initializing pub-sub from either config file or cli. Refactored existing tests to use fixture.

xlcheng1 (approved):

HarshGandhi-AWS (approved):
LGTM

Issue #175, opened 2021-10-15
Logging target override doesn't change SDK logging


rpcme:
**Describe the bug**

Invoked device client with config file, which has the following configuration

```json
	"logging": {
		"level": "DEBUG",
		"type": "FILE",
		"file": "/var/log/aws-iot-device-client.log"
	},

```

Output of SDK log is in a different directory

```text
2021-08-04T15:14:52.711Z [DEBUG] {Config.cpp}: Did not find a runtime configuration file, assuming Fleet Provisioning has not run for this device
2021-08-04T15:14:52.711Z [INFO]  {Main.cpp}: Now running AWS IoT Device Client version v1.4.0
2021-08-04T15:14:52.717Z [INFO]  {SharedCrtResourceManager.cpp}: SDK logging is enabled. Check /var/log/aws-iot-device-client/sdk.log for SDK logs.
2021-08-04T15:14:52.718Z [DEBUG] {Retry.cpp}: Retryable function starting, it will retry until success
```

**To Reproduce**

1. run device client with different logging target in configuration

**Expected behavior**

both aws iot device client and sdk log are in the same directory

**Actual behavior**

aws iot device client log is in /var/log, but sdk log is in different
directory as shown in the aws iot device client log output

**Logs**

If applicable, add full logs of errors and outputs to help explain your problem. Preferably, you can also [increase the verbosity of the logs](https://github.com/awslabs/aws-iot-device-client#logging).

**Environment (please complete the following information):**
 - OS: Linux
 - Version 1.4.0
 - Architecture: x86-64, RISC-V
 - Device Client commit: 1.4.0 tag

**Additional context**

HarshGandhi-AWS:
Hello @rpcme , again thank you for reaching out to us with this issue. @marcoemorais-aws has released the [bug fix](https://github.com/awslabs/aws-iot-device-client/pull/199) for this issue. Please review and test it to see if the issue is resolved for you or not. 
I am resolving this issue now, please feel free to reopen it if the issue is not resolved or open a new issue with us if you are facing any difficulty using the Device Client.

Regards, 
Harsh Gandhi

PR #199, opened 2021-12-08
Setup SDK logging from Config as part of CRT SDK initialization

marcoemorais-aws:
### Motivation
- The SDK logging feature has a few unexpected misbehaviors.
  - If the destination SDK log file does not exist, then SDK logging will silently fail.
  - SDK logging does not enforce the same permissions requirements as device client log.
- Issue number: https://github.com/awslabs/aws-iot-device-client/issues/175

#### Change summary
- Add logic to create parent directories and file with expected permissions as part of CRT SDK initialization.
  - This logic is functionally equivalent to behavior of device client log file.
- Add SDK logging configuration to the output of `--export-default-settings`.
- Fix bug in `FileUtils::CreateDirectoryWithPermissions` in return value when desired and actual file permissions match. 

#### Revision diff summary

### Testing
* Tested the behavior of SDK logging with the following permutations.
  * SDK logging options set from the command line and from config file.
  * SDK logging configured with file paths containing directories and files that do not exist.
* Add unit tests to `FileUtlis` for the new functions used in this feature.
* We currently do not have tests for `SharedCrtResourceManager`, this should be added in a future commit.

### DIFF
--- a/source/config/Config.cpp
+++ b/source/config/Config.cpp
@@ -344,8 +344,6 @@ constexpr char PlainConfig::LogConfig::JSON_KEY_ENABLE_SDK_LOGGING[];
 constexpr char PlainConfig::LogConfig::JSON_KEY_SDK_LOG_LEVEL[];
 constexpr char PlainConfig::LogConfig::JSON_KEY_SDK_LOG_FILE[];
 
-constexpr char PlainConfig::LogConfig::DEFAULT_SDK_LOG_FILE[];
-
 int PlainConfig::LogConfig::ParseDeviceClientLogLevel(string level)
 {
     string temp = level;
@@ -1763,7 +1761,10 @@ bool Config::ExportDefaultSetting(const string &file)
     "%s": {
         "%s": "DEBUG",
         "%s": "FILE",
-        "%s": "/var/log/aws-iot-device-client/aws-iot-device-client.log"
+        "%s": "/var/log/aws-iot-device-client/aws-iot-device-client.log",
+        "%s": false,
+        "%s": "TRACE",
+        "%s": "/var/log/aws-iot-device-client/sdk.log"
     },
     "%s": {
         "%s": true,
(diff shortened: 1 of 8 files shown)


source/config/Config.cpp
```diff
@@ -1763,7 +1761,10 @@ bool Config::ExportDefaultSetting(const string &file)
     "%s": {
         "%s": "DEBUG",
         "%s": "FILE",
-        "%s": "/var/log/aws-iot-device-client/aws-iot-device-client.log"
+        "%s": "/var/log/aws-iot-device-client/aws-iot-device-client.log",
+        "%s": false,
+        "%s": "TRACE",
+        "%s": "/var/log/aws-iot-device-client/sdk.log"
```
HarshGandhi-AWS, line 1767: We can use constants storing default values over here and above.

source/util/FileUtils.cpp
```diff
@@ -296,6 +303,35 @@ bool FileUtils::CreateDirectoryWithPermissions(const char *dirPath, mode_t permi
     return false;
 }
 
+bool FileUtils::DirectoryExists(const std::string &dirPath)
+{
+    struct stat dirInfo;
+    if (stat(dirPath.c_str(), &dirInfo) != 0)
```
HarshGandhi-AWS, line 309: I think we should extract the expanded absolute path of the directory before calling `stat` on the directory.

source/util/FileUtils.cpp
```diff
@@ -296,6 +303,35 @@ bool FileUtils::CreateDirectoryWithPermissions(const char *dirPath, mode_t permi
...
+    {
+        // Ignore permission errors.
+        return false;
+    }
+    return S_ISDIR(dirInfo.st_mode);
+}
+
+bool FileUtils::CreateEmptyFileWithPermissions(const string &filename, mode_t permissions)
+{
+    auto fd = open(filename.c_str(), O_CREAT | O_EXCL, permissions);
```
HarshGandhi-AWS, line 319: I think we should extract the expanded absolute path of the file before calling `open` to create one.

source/config/Config.cpp
marcoemorais-aws: Done.

source/util/FileUtils.cpp
marcoemorais-aws: Done.

source/util/FileUtils.cpp
marcoemorais-aws: Done.

xlcheng1 (approved):

HarshGandhi-AWS (approved):
LGTM

PR #204, opened 2021-12-13
enabled dir creation in setup.sh and dir/file creation in pubsub.cpp

xlcheng1:
### Motivation
- Added PubSub directory creation in setup.sh. Added PubSub Directory/files creation in pubsub.cpp

#### Change summary

#### Revision diff summary

### Testing
- CI test run result: <link>

### DIFF
--- a/source/samples/pubsub/PubSubFeature.cpp
+++ b/source/samples/pubsub/PubSubFeature.cpp
@@ -3,11 +3,13 @@
 
 #include "PubSubFeature.h"
 #include "../../logging/LoggerFactory.h"
+#include "../../util/FileUtils.h"
 
 #include <aws/common/byte_buf.h>
 #include <aws/crt/Api.h>
 #include <aws/iotdevicecommon/IotDevice.h>
 #include <iostream>
+#include <sys/stat.h>
 
 #include <utility>
 
@@ -22,12 +24,62 @@ using namespace Aws::Iot::DeviceClient::Util;
 using namespace Aws::Iot::DeviceClient::Logging;
 
 constexpr char PubSubFeature::TAG[];
+constexpr char PubSubFeature::DEFAULT_PUBLISH_FILE[];
+constexpr char PubSubFeature::DEFAULT_SUBSCRIBE_FILE[];
 
 #define MAX_IOT_CORE_MQTT_MESSAGE_SIZE_BYTES 128000
 
 string PubSubFeature::getName()
 {
     return "Pub Sub Sample";
+}
+
+bool PubSubFeature::createPubSub(const PlainConfig &config, std::string absFilePath)
+{
+    std::string pubFileDir = FileUtils::ExtractParentDirectory(absFilePath);
+    if (!FileUtils::DirectoryExists(pubFileDir))
+    {
+        // Create an empty directory with the expected permissions.
+        FileUtils::CreateDirectoryWithPermissions(pubFileDir.c_str(), S_IRWXU | S_IRGRP | S_IROTH | S_IXOTH);
+    }
+    else
+    {
+        // Verify the directory permissions.
+        auto rcvDirPermissions = FileUtils::GetFilePermissions(pubFileDir);
+        if (Permissions::PUBSUB_DIR != rcvDirPermissions)
+        {
+            LOGM_ERROR(
+                TAG,
+                "Incorrect directory permissions for pubsub file: %s expected: %d received: %d",
+                Sanitize(pubFileDir).c_str(),
+                Permissions::PUBSUB_DIR,
+                rcvDirPermissions);
+            return false;
+        }
+    }
+
+    if (!FileUtils::FileExists(absFilePath))
+    {
+        // Create an empty file with the expected permissions.
+        FileUtils::CreateEmptyFileWithPermissions(absFilePath, S_IRUSR | S_IWUSR);
+    }
+    else
+    {
+        // Verify the file permissions.
+        auto rcvFilePermissions = FileUtils::GetFilePermissions(absFilePath);
+        if (Permissions::PUB_SUB_FILES != rcvFilePermissions)
+        {
+            LOGM_ERROR(
+                TAG,
+                "Incorrect file permissions for pubsub file: %s expected: %d received: %d",
+                Sanitize(absFilePath).c_str(),
+                Permissions::PUB_SUB_FILES,
+                rcvFilePermissions);
+            return false;
+        }
+    }
+
+    return true;
 }
 
 int PubSubFeature::init(
@@ -40,8 +92,26 @@ int PubSubFeature::init(
     thingName = *config.thingName;
     pubTopic = config.pubSub.publishTopic.value();
     subTopic = config.pubSub.subscribeTopic.value();
-    pubFile = config.pubSub.publishFile.has_value() ? config.pubSub.publishFile.value() : "";
-    subFile = config.pubSub.subscribeFile.has_value() ? config.pubSub.subscribeFile.value() : "";
+
+    if (!config.pubSub.publishFile->empty())
+    {
+        pubFile = config.pubSub.publishFile->c_str();
+    }
+
+    if (!createPubSub(config, pubFile))
+    {
+        LOG_ERROR(TAG, "Failed to create publish directory or file");
+    }
+
+    if (!config.pubSub.subscribeFile->empty())
+    {
+        subFile = config.pubSub.subscribeFile->c_str();
+    }
+
+    if (!createPubSub(config, subFile))
+    {
+        LOG_ERROR(TAG, "Failed to create subscribe directory or file");
+    }
 
     return AWS_OP_SUCCESS;
 }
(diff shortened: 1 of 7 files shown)


source/samples/pubsub/PubSubFeature.cpp
```diff
@@ -83,7 +153,8 @@ void PubSubFeature::publishFileData()
         LOG_ERROR(TAG, "Failed to read publish file... Skipping publish");
         return;
     }
-    auto onPublishComplete = [payload, this](Mqtt::MqttConnection &, uint16_t packetId, int errorCode) mutable {
+    auto onPublishComplete = [payload, this](Mqtt::MqttConnection &, uint16_t packetId, int errorCode) mutable
+    {
         LOGM_DEBUG(TAG, "PublishCompAck: PacketId:(%u), ErrorCode:%i", getName().c_str(), errorCode);
```
marcoemorais-aws, line 158: There seems to be a mismatch in logging format strings.
* `%u` - this is `unsigned int`, but the type of `getName().c_str()` is a null-terminated string, suggest change to `%s`
* alternatively if the intent was to log the `packetId`, then replace `getName().c_str()` with `packetId` in the first position
 

https://man7.org/linux/man-pages/man3/printf.3.html

source/samples/pubsub/PubSubFeature.cpp
```diff
@@ -96,10 +167,10 @@ int PubSubFeature::start()
     LOGM_INFO(TAG, "Starting %s", getName().c_str());
 
     auto onSubAck =
-        [&](MqttConnection &connection, uint16_t packetId, const String &topic, QOS qos, int errorCode) -> void {
-        LOGM_DEBUG(TAG, "SubAck: PacketId:(%u), ErrorCode:%i", getName().c_str(), errorCode);
```
marcoemorais-aws, line 100: Same comment as above, format string mismatch.

source/samples/pubsub/PubSubFeature.cpp
```diff
@@ -96,10 +167,10 @@ int PubSubFeature::start()
...
-    };
-    auto onRecvData = [&](MqttConnection &connection, const String &topic, const ByteBuf &payload) -> void {
+        [&](MqttConnection &connection, uint16_t packetId, const String &topic, QOS qos, int errorCode) -> void
+    { LOGM_DEBUG(TAG, "SubAck: PacketId:(%u), ErrorCode:%i", getName().c_str(), errorCode); };
+    auto onRecvData = [&](MqttConnection &connection, const String &topic, const ByteBuf &payload) -> void
+    {
         LOGM_DEBUG(TAG, "Message received on subscribe topic, size: %zu bytes", payload.len);
         if (string((char *)payload.buffer) == PUBLISH_TRIGGER_PAYLOAD)
```
marcoemorais-aws, line 175: This code assumes that `payload.buffer` points to a null-terminated string. That is a strong assumption and, if not true, then you have a potential for buffer overflow. It will be more secure to use the 2-argument constructor to `std::string` that takes a pointer to character (`payload.buffer`) and number of characters (`payload.len`).  See `std::string` constructor labeled as (4) in the doc below.

https://en.cppreference.com/w/cpp/string/basic_string/basic_string

source/samples/pubsub/PubSubFeature.cpp
```diff
@@ -30,6 +34,54 @@ string PubSubFeature::getName()
     return "Pub Sub Sample";
 }
 
+bool PubSubFeature::createPubSub(const PlainConfig &config, std::string absFilePath)
+{
+    std::string pubFileDir = FileUtils::ExtractParentDirectory(absFilePath);
+    if (!FileUtils::DirectoryExists(pubFileDir))
+    {
+        // Create an empty directory with the expected permissions.
+        FileUtils::CreateDirectoryWithPermissions(pubFileDir.c_str(), S_IRWXU | S_IRGRP | S_IROTH | S_IXOTH);
```
marcoemorais-aws, line 43: This function returns `true/false` on `success/failure`. You have to test for return and propagate back to caller.

```
if (!FileUtils::CreateDirectoryWithPermissions(pubFileDir.c_str(), S_IRWXU | S_IRGRP | S_IROTH | S_IXOTH))
{
    return false;
}
```

source/samples/pubsub/PubSubFeature.cpp
```diff
@@ -30,6 +34,54 @@ string PubSubFeature::getName()
...
+                Permissions::PUBSUB_DIR,
+                rcvDirPermissions);
+            return false;
+        }
+    }
+
+    if (!FileUtils::FileExists(absFilePath))
+    {
+        // Create an empty file with the expected permissions.
+        FileUtils::CreateEmptyFileWithPermissions(absFilePath, S_IRUSR | S_IWUSR);
```
marcoemorais-aws, line 64: This function returns `true/false` on `success/failure`. You have to test for return and propagate back to caller.

```
if (!FileUtils::CreateEmptyFileWithPermissions(absFilePath, S_IRUSR | S_IWUSR))
{
    return false;
}
```

source/samples/pubsub/PubSubFeature.cpp
```diff
@@ -30,6 +34,54 @@ string PubSubFeature::getName()
...
+            LOGM_ERROR(
+                TAG,
+                "Incorrect file permissions for pubsub file: %s expected: %d received: %d",
+                Sanitize(absFilePath).c_str(),
+                Permissions::PUB_SUB_FILES,
+                rcvFilePermissions);
+            return false;
+        }
+    }
+
```
marcoemorais-aws, line 81: [Minor] Suggest to include a log message at INFO that prints pubsub file path.

source/samples/pubsub/PubSubFeature.cpp
```diff
@@ -40,8 +92,26 @@ int PubSubFeature::init(
     thingName = *config.thingName;
     pubTopic = config.pubSub.publishTopic.value();
     subTopic = config.pubSub.subscribeTopic.value();
-    pubFile = config.pubSub.publishFile.has_value() ? config.pubSub.publishFile.value() : "";
-    subFile = config.pubSub.subscribeFile.has_value() ? config.pubSub.subscribeFile.value() : "";
+
+    if (!config.pubSub.publishFile->empty())
+    {
+        pubFile = config.pubSub.publishFile->c_str();
```
marcoemorais-aws, line 98: Both `Config::PubSub::publishFile` and `Config::PubSub::subscribeFile` are of type `Aws::Crt::Optional<std::string>`. This means you need 2 tests:
* `has_value()` to confirm that the optional value is set
* `empty()` on the value to confirm that the value (`std::string`) is not empty

I won't repeat but same thing applies a few lines below for the `subscribeFile`.

source/samples/pubsub/PubSubFeature.cpp
xlcheng1: done

source/samples/pubsub/PubSubFeature.cpp
xlcheng1: done

source/samples/pubsub/PubSubFeature.cpp
xlcheng1: done

source/samples/pubsub/PubSubFeature.cpp
xlcheng1: done

source/samples/pubsub/PubSubFeature.cpp
xlcheng1: done

source/samples/pubsub/PubSubFeature.cpp
xlcheng1: done

marcoemorais-aws (approved):

HarshGandhi-AWS (approved):
LGTM

source/samples/pubsub/PubSubFeature.cpp
```diff
@@ -40,8 +99,28 @@ int PubSubFeature::init(
...
+
+    if (config.pubSub.publishFile.has_value() && !config.pubSub.publishFile->empty())
+    {
+        pubFile = config.pubSub.publishFile->c_str();
+    }
+    pubFile = FileUtils::ExtractExpandedPath(pubFile);
+
+    if (!createPubSub(config, pubFile))
+    {
+        LOG_ERROR(TAG, "Failed to create publish directory or file");
```
marcoemorais-aws, line 111: @xlcheng1 Failure to create the pub sub files PubSubFeature::init is not being propagated back to caller, I think that should be changed. Same applies below.  What do you think @HarshGandhi-AWS?

HarshGandhi-AWS: In DC right now we are not considering failures in sample as Fatal error. We should consider changing this behavior.

source/samples/pubsub/PubSubFeature.h
```diff
@@ -73,16 +77,15 @@ namespace Aws
                     /**
                      * \brief Location of file containing data to publish
                      */
-                    std::string pubFile;
+                    std::string pubFile = DEFAULT_PUBLISH_FILE;
```
HarshGandhi-AWS, line 80: Think this code should be moved to Config.cpp and Config.h so that we do not override the expanded path. Also by doing this we won't have to call expanded path function multiple times.

source/samples/pubsub/PubSubFeature.cpp
xlcheng1: will address it in a separate PR

source/samples/pubsub/PubSubFeature.h
xlcheng1: will address it in a separate PR

Issue #211, opened 2022-01-16
Building demos with the AWS IoT Device Client FAILS


Sriram-Y:
**Describe the bug**
The Run AWS IoT Device Client step as described in:
https://docs.aws.amazon.com/iot/latest/developerguide/iot-dc-install-configure.html

Fails to publish the "Hello World!" MQTT message in MQTT test client dashboard.

**To Reproduce**
1. Go through all the steps given in https://docs.aws.amazon.com/iot/latest/developerguide/iot-dc-install-configure.html

**Expected behavior**
"Hello World!" should be display in the MQTT test client dashboard.

**Actual behavior**
No message displayed in MQTT test client dashboard.
See error log on device side.

**Logs**
2022-01-16T01:05:28.614Z [INFO]  {samples/PubSubFeature.cpp}: Creating Pub/Sub file: /home/pi/.aws-iot-device-client/pubsub/publish-file.txt 
2022-01-16T01:05:28.614Z [INFO]  {samples/PubSubFeature.cpp}: Creating Pub/Sub file: /home/pi/.aws-iot-device-client/pubsub/subscribe-file.txt 
2022-01-16T01:05:28.614Z [INFO]  {SharedCrtResourceManager.cpp}: Starting Device Client features. 
2022-01-16T01:05:28.614Z [INFO]  {samples/PubSubFeature.cpp}: Starting Pub Sub Sample 
2022-01-16T01:05:28.614Z [ERROR] {samples/PubSubFeature.cpp}: Publish file contains no data 
2022-01-16T01:05:28.614Z [ERROR] {samples/PubSubFeature.cpp}: Failed to read publish file... Skipping publish 
2022-01-16T01:05:28.614Z [INFO]  {Main.cpp}: Client base has been notified that Pub Sub Sample has started 
2022-01-16T01:05:28.665Z [DEBUG] {samples/PubSubFeature.cpp}: SubAck: PacketId:(Pub Sub Sample), ErrorCode:0 
2022-01-16T01:05:49.296Z [INFO]  {Main.cpp}: Received signal: (2) 
2022-01-16T01:05:49.296Z [DEBUG] {Main.cpp}: Attempting shutdown of Pub Sub Sample 
2022-01-16T01:05:49.297Z [INFO]  {Main.cpp}: Pub Sub Sample has stopped 
2022-01-16T01:05:49.297Z [INFO]  {Main.cpp}: All features have stopped 
[...]

PR #212, opened 2022-01-16
Fix for publish of DEFAULT_PUBLISH_PAYLOAD

Sriram-Y:
When AWS IoT Device Client is run on a clean install, the PubSub test for "Hello World!" fails.
Fixing it.

### Motivation
- Issue number: 211
(https://github.com/awslabs/aws-iot-device-client/issues/211)

### Modifications
Line 155 in aws-iot-device-client/source/samples/pubsub/PubSubFeature.cpp publishFileData() fixed.

#### Change summary
See above.

#### Revision diff summary
N/A

### Testing
Tested with and without publish-file.txt file.
Also tested with and without content in the file.
In all 4 cases either "Hello World!" was received or custom message was received in the MQTT test client.

### DIFF
--- a/source/samples/pubsub/PubSubFeature.cpp
+++ b/source/samples/pubsub/PubSubFeature.cpp
@@ -152,7 +152,7 @@ int PubSubFeature::getPublishFileData(aws_byte_buf *buf)
 void PubSubFeature::publishFileData()
 {
     ByteBuf payload;
-    if (pubFile == "")
+    if (FileUtils::GetFileSize(pubFile) == 0)
     {
         aws_byte_buf_init(&payload, resourceManager->getAllocator(), DEFAULT_PUBLISH_PAYLOAD.size());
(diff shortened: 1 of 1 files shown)


source/samples/pubsub/PubSubFeature.cpp
```diff
@@ -152,7 +152,7 @@ int PubSubFeature::getPublishFileData(aws_byte_buf *buf)
 void PubSubFeature::publishFileData()
 {
     ByteBuf payload;
-    if (pubFile == "")
+    if (FileUtils::GetFileSize(pubFile) == 0)
```
HarshGandhi-AWS, line 155: Are we adding default message in publish file while creating it? If not then the issue will not solve the problem of our new DC users who has not manually update the file data.

HarshGandhi-AWS:
Hello @Sriram-Y  , since we haven't heard back from you for a while, we are closing this review for now. Feel free to reopen this review or create a new one if you find any issue using Device Client. Thank you!

PR #215, opened 2022-02-07
Attempt to locate OpenSSL on MacOS if pkg-config detection fails

timmattison:
### Motivation
- On MacOS OpenSSL may not be found by pkg-config. If it is not found the user then has to specify the OpenSSL root directory which they may not know. Since many Mac users use homebrew this patch attempts to use pkg-config to find OpenSSL and then falls back on the Cellar directory.

#### Change summary
One modification to the CMakeLists.txt file that will search for homebrew installations of OpenSSL (on MacOS only) and then find the latest version to build with.

### Testing
No CI for this but I needed to make this change in order to build the project on Mac OS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,26 @@ configure_file("source/Version.h.in" "${PROJECT_BINARY_DIR}/Version.h")
 # OpenSSL dependency                    #
 #########################################
 set(OPENSSL_USE_STATIC_LIBS TRUE)
+
+if(APPLE)
+    # Try to find OpenSSL on MacOS but don't fail immediately so we can fall back to alternate logic
+    find_package(OpenSSL)
+
+    IF(NOT OpenSSL_FOUND)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -133,6 +133,26 @@ configure_file("source/Version.h.in" "${PROJECT_BINARY_DIR}/Version.h")
 # OpenSSL dependency                    #
 #########################################
 set(OPENSSL_USE_STATIC_LIBS TRUE)
+
+if(APPLE)
```
HarshGandhi-AWS, line 137: At this moment, Device Client does not provide support for MacOS. This is mainly because of the shadow feature of device client which does not work on MacOS because of dependency issue which is expected and known. Even if your change over here solves the issue of finding the OpenSSL on MacOS, we cannot push this change since it will give the wrong impression of device client is being supported on MacOS which is not true. For this reason we will have to close this review. 

When we add support for Device Client on MacOS, we can reopen this review and merge your changes. 

Thank you.

PR #217, opened 2022-02-07
Switch to fork from vfork

timmattison:
### Motivation
- The compilation phase on MacOS treats usage of vfork as an error instead of a warning. This changes vfork -> fork to allow compilation on MacOS.
-  
#### Change summary
Just one change of vfork -> fork

### Testing
No CI for this but I needed to make this change in order to build the project on Mac OS. fork does work slightly differently than vfork though so I may add some more comments later once I've verified that it works as expected.

### DIFF
--- a/source/jobs/JobEngine.cpp
+++ b/source/jobs/JobEngine.cpp
@@ -270,7 +270,7 @@ int JobEngine::exec_cmd(string operation, PlainJobDocument::JobAction action)
 
     int execResult;
     int returnCode;
-    int pid = vfork();
+    int pid = fork();
     if (pid < 0)
     {
         LOGM_ERROR(TAG, "Failed to create child process, fork returned %d", pid);


monelgordillo:
During "cmake --build . --target aws-iot-device-client" I get this:

```
aws-iot-device-client/source/jobs/JobEngine.cpp:276:15: error: 'vfork' is deprecated: Use posix_spawn or fork [-Werror,-Wdeprecated-declarations]
    int pid = vfork();
```

source/jobs/JobEngine.cpp
```diff
@@ -270,7 +270,7 @@ int JobEngine::exec_cmd(string operation, PlainJobDocument::JobAction action)
 
     int execResult;
     int returnCode;
-    int pid = vfork();
+    int pid = fork();
```
HarshGandhi-AWS, line 273: Child process created by`vfork()` suspends execution of parent process until child process completes its execution  wereas, both parent and the child process created using`fork()` start their execution from next statement after fork() and both processes get executed simultaneously.

This might result in executing multiple job executions simultaneously. 

Device Client anyways does not work on Mac OS so I would not encourage making this change until we make the software MacOS compatible.

timmattison: The child and parent process do different things. The parent process waits for the child process to finish already.

HarshGandhi-AWS: I might have had a wrong understanding then. @timmattison let me test these changes locally and review this change again.

HarshGandhi-AWS (approved):

PR #218, opened 2022-02-07
Skip shadow examples on MacOS due to inotify not being supported

timmattison:
### Motivation
- MacOS does not support inotify but the shadow example code uses it. On MacOS these examples need to be removed in order for the compilation to be successful.
  
#### Change summary
Added a check in CMakeLists.txt to detect if the system is an Apple operating system. If so, disable the shadow examples.

### Testing
No CI for this but I needed to make this change in order to build the project on Mac OS.


HarshGandhi-AWS:
At this moment, Device Client does not provide support for MacOS. This is mainly because of the shadow feature of device client which does not work on MacOS because of dependency issue which is expected and known. Even if your change over here disables Shadow feature on MacOS, we cannot push this change since it will give the wrong impression of device client is being supported on MacOS which is not true. For this reason we will have to close this review.

When we add support for Device Client on MacOS, we can reopen this review and merge your changes.

Thank you.

PR #225, opened 2022-03-09
Add logging option enabling both STDOUT and FILE output

BolongZhang-AWS:
### Motivation
Add logging option enabling both STDOUT and FILE output

#### Change summary
Added StdOutFileLogger

#### Revision diff summary

### Testing
Build
- CI test run result: <link>

### DIFF
--- a/source/logging/StdOutFileLogger.cpp
+++ b/source/logging/StdOutFileLogger.cpp
@@ -125,8 +125,8 @@ void StdOutFileLogger::writeLogMessage(unique_ptr<LogMessage> message)
 
     *outputStream << time_buffer << " " << LogLevelMarshaller::ToString(message->getLevel()) << " {"
                   << message->getTag() << "}: " << message->getMessage() << endl;
-    cout << time_buffer << " " << LogLevelMarshaller::ToString(message->getLevel()) << " {"
-         << message->getTag() << "}: " << message->getMessage() << endl;
+    cout << time_buffer << " " << LogLevelMarshaller::ToString(message->getLevel()) << " {" << message->getTag()
+         << "}: " << message->getMessage() << endl;
     outputStream->flush();
 }
 
@@ -181,5 +181,3 @@ void StdOutFileLogger::flush()
         }
     }
 }
-
-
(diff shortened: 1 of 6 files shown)


HarshGandhi-AWS:
Make sure we update our documentation. 

https://github.com/awslabs/aws-iot-device-client/tree/main/source/logging#logging

source/logging/StdOutFileLogger.h
```diff
@@ -0,0 +1,119 @@
...
+namespace Aws
+{
+    namespace Iot
+    {
+        namespace DeviceClient
+        {
+            namespace Logging
+            {
+                /**
+                 * \brief File-based logging implementation for writing log messages to a file on the device
```
HarshGandhi-AWS, line 25: Lets update this comment with correct message.

source/logging/StdOutFileLogger.h
```diff
@@ -0,0 +1,119 @@
...
+                     *
+                     * This method will write the log message to the file specified for logging
+                     * @param message the message to log
+                     */
+                    void writeLogMessage(std::unique_ptr<LogMessage> message);
+
+                    /**
+                     * \brief Creates the directories required as part of the full path to the desired log file
+                     *
+                     * This method is run as part of the initialization process of the FileLogger implementation
```
HarshGandhi-AWS, line 70: This line of comment also need to be updated.

source/logging/StdOutFileLogger.cpp
```diff
@@ -0,0 +1,183 @@
...
+void StdOutFileLogger::stop()
+{
+    needsShutdown = true;
+    logQueue->shutdown();
+
+    unique_lock<mutex> runLock(isRunningLock);
+    isRunning = false;
+}
+
+void StdOutFileLogger::writeLogMessage(unique_ptr<LogMessage> message)
```
HarshGandhi-AWS, line 121: Except adding logic to write logs to a file and write in `cout` statement, everything written in this file is same as `FileLogger.cpp` and `StdOutLogger.cpp`. Can we instead update the structure of logger feature to avoid this redundancy? instead of creating these three classes, we can only create one single class for `Logger` and add different methods to write logs in file and/or Standard output.

HarshGandhi-AWS (changes requested):

PR #227, opened 2022-03-16
Adds additional check for package

shangabl:
### Motivation

The handler was failing in the event that both dpkg and rpm were both installed

#### Change summary

The verify-packages-installed handler will fail if both rpm and dpkg are installed. This fixes that.

#### Revision diff summary

### Testing
Passed integration tests successfully.

### DIFF
--- a/sample-job-handlers/verify-packages-installed.sh
+++ b/sample-job-handlers/verify-packages-installed.sh
@@ -8,28 +8,29 @@ packages=$@
 echo "Username: $user"
 echo "Packages to verify: $packages"
 
-if command -v "rpm" > /dev/null; then
+if id "$user" 2>/dev/null && command -v "sudo" > /dev/null; then
   for package in $packages
   do
-    if id "$user" 2>/dev/null && command -v "sudo" > /dev/null; then
-      sudo -u "$user" -n rpm -q "$package"
-    else
-      echo "username or sudo command not found"
-      rpm -q "$package"
+    if command -v "rpm" >/dev/null; then
+      if command sudo -u "$user" -n rpm -q "$package" > /dev/null; then
+        continue
+      fi
     fi
-  done
-elif command -v "dpkg" > /dev/null; then
-  # Using -s flag instead of -l for dpkg based on https://github.com/bitrise-io/bitrise/issues/433
-  for package in $packages
-  do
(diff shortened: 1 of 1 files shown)


HarshGandhi-AWS:
How can it fail if both `dpkg` and `rpm` were installed on the machine? We have the `if-elseif` condtional statements to check that?

sample-job-handlers/verify-packages-installed.sh
```diff
@@ -8,28 +8,29 @@ packages=$@
...
+if id "$user" 2>/dev/null && command -v "sudo" > /dev/null; then
   for package in $packages
   do
-    if id "$user" 2>/dev/null && command -v "sudo" > /dev/null; then
-      sudo -u "$user" -n rpm -q "$package"
-    else
-      echo "username or sudo command not found"
-      rpm -q "$package"
+    if command -v "rpm" >/dev/null; then
+      if command sudo -u "$user" -n rpm -q "$package" > /dev/null; then
```
HarshGandhi-AWS, line 15: why do we have a conditional check over here? we are not doing anything if it returns either `true` or `false`.

sample-job-handlers/verify-packages-installed.sh
```diff
@@ -8,28 +8,29 @@ packages=$@
...
+    if command -v "dpkg" >/dev/null; then
       dpkg -s "$package"
     fi
   done
-else
-  >&2 echo "No suitable package manager found"
-  exit 1
-fi;
\ No newline at end of file
+fi;
```
HarshGandhi-AWS, line 36: what if both `dpkg` and `rpm` is not installed? We should exit with status code 1 in that case? Why did we remove that logic?

HarshGandhi-AWS (changes requested):

sample-job-handlers/verify-packages-installed.sh
shangabl: If true we move on to the next package, if false we check dpkg. I did forgot to exit 1 though if both fail... will fix.

sample-job-handlers/verify-packages-installed.sh
shangabl: Missed this as well... As for your other comment if both were installed it was only checking one of them because of the elif. Will fix

PR #229, opened 2022-03-21
Adds additional checks for packages

shangabl:
### Motivation
Integration tests fail if multiple package managers installed

#### Change summary
Modifies verify packages installed script to check for packages with both rpm and dpkg.

### Testing
Ran Jobs integration tests

### DIFF
--- a/sample-job-handlers/verify-packages-installed.sh
+++ b/sample-job-handlers/verify-packages-installed.sh
@@ -8,28 +8,41 @@ packages=$@
 echo "Username: $user"
 echo "Packages to verify: $packages"
 
-if command -v "rpm" > /dev/null; then
+if id "$user" 2>/dev/null && command -v "sudo" > /dev/null; then
   for package in $packages
   do
-    if id "$user" 2>/dev/null && command -v "sudo" > /dev/null; then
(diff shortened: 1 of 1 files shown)


sample-job-handlers/verify-packages-installed.sh
```diff
@@ -8,28 +8,41 @@ packages=$@
...
+if id "$user" 2>/dev/null && command -v "sudo" > /dev/null; then
   for package in $packages
   do
-    if id "$user" 2>/dev/null && command -v "sudo" > /dev/null; then
-      sudo -u "$user" -n rpm -q "$package"
-    else
-      echo "username or sudo command not found"
-      rpm -q "$package"
+    if command -v "rpm" >/dev/null; then
+      if command sudo -u "$user" -n rpm -q "$package" > /dev/null; then
```
BolongZhang-AWS, line 15: There's no else case if the command return not null? seems we do "exit 1" for the dpkg case.

shangabl: It needs to check both rpm & dpkg and if the package isn't present for both then we exit 1. So, we can't exit 1 after we check rpm. I can remove the else from the dpkg check as I suppose it is redundant.

PR #235, opened 2022-04-04
Adds Permissions info to ST ReadMe

shangabl:
### Motivation
We were lacking an explanation of the necessary permissions to use secure tunneling.

#### Change summary
Adds details about the required permissions to the ReadMe.

### DIFF
--- a/source/shadow/README.md
+++ b/source/shadow/README.md
@@ -48,7 +48,8 @@ Device Client would also perform configuration validation check when detecting t
 ### Policy Permissions
 In order to use the Shadow feature the device must first have permission to connect to IoT Core.
 The device must also be able to publish, subscribe, and receive messages on the Shadow topics. You can read more [here](https://docs.aws.amazon.com/iot/latest/developerguide/device-shadow-mqtt.html).
-The example policy below demonstrates the minimum permissions required. Simply replace the `<region>` and `<accountId`> with the correct values.
+The example policy below demonstrates the minimum permissions required for the Named Shadow feature. 
+Replace the `<region>`, `<accountId`> and `<shadowName>` with the correct values.
 ```
 {
   "Version": "2012-10-17",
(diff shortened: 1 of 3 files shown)


source/tunneling/README.md
```diff
@@ -55,6 +56,34 @@ $ ./localproxy -r us-east-1 -s 8080 -t <source token>
 $ ssh -p 8080 <remote_user_name>@localhost
 ```
 
+### Policy Permissions
+In order to use the Secure Tunneling feature the device must first have permission to connect to IoT Core. 
+The device must also be able to subscribe to and receive messages on the Secure Tunneling notify topic in order to receive the destination token.
+The example policy below demonstrates the minimum permissions required. Simply replace the `<region>` and `<accountId`> with the correct values.
```
HarshGandhi-AWS, line 62: Can you update this line to -> `The example policy below demonstrates the minimum permissions required to run Secure Tunneling feature.`

HarshGandhi-AWS (commented):
Let's also add sample policy in other IoT feature `Readme`. It can be done in a separate PR if required.

source/shadow/README.md
```diff
@@ -45,4 +45,130 @@ Device Client would also perform configuration validation check when detecting t
...
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "iot:Receive"
+      ],
+      "Resource": [
+        "arn:aws:iot:<region>:<account>:topic/$aws/things/${iot:Connection.Thing.ThingName}/shadow/update/rejected"
+      ]
+    }
```
HarshGandhi-AWS, line 152: `,` is missing in this policy. Can you verify that all other policies are correct? It would be better if we test those.  Also we have a size limit for policies so some of these might be above the size limit of the policy file.

source/shadow/README.md
```diff
@@ -45,4 +45,130 @@ Device Client would also perform configuration validation check when detecting t
 ```
 *It is important to note that providing input and output files is optional.  If they are not configured or left blank, "{welcome","aws-iot}" will be published to this shadow and the `default-sample-shadow-document`file will be created under this path `~/.aws-iot-device-client/sample-shadow/` by default*
 
+### Policy Permissions
+In order to use the Shadow feature the device must first have permission to connect to IoT Core.
+The device must also be able to publish, subscribe, and receive messages on the Shadow topics. You can read more [here](https://docs.aws.amazon.com/iot/latest/developerguide/device-shadow-mqtt.html).
+The example policy below demonstrates the minimum permissions required. Simply replace the `<region>` and `<accountId`> with the correct values.
```
HarshGandhi-AWS, line 51: update this line `the minimum permissions required` to `the minimum permissions required for Shadow feature.`

source/shadow/README.md
```diff
@@ -45,4 +45,130 @@ Device Client would also perform configuration validation check when detecting t
...
+      "Action": [
+        "iot:Receive"
+      ],
+      "Resource": [
+        "arn:aws:iot:<region>:<account>:topic/$aws/things/${iot:Connection.Thing.ThingName}/shadow/get/accepted"
+      ]
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
```
HarshGandhi-AWS, line 92: Consider combining these policy sections with same `Action` to reduce the file size.

shangabl: Okay yeah makes sense... I was thinking this might be better but they are getting pretty big.

shangabl:
Fully tested these example polices.

HarshGandhi-AWS (approved):
LGTM

marcoemorais-aws (approved):

PR #236, opened 2022-04-05
Prevent multiple processes

RogerZhongAWS:
#### Change summary
Prevents multiple Device Client processes from existing at the same time. Creates a lock file upon process start, deletes it upon process end. NOTE: Writing to /var/run requires sudo permissions, device client will seg fault if not run with sudo. Please advise whether using this directory for writing the lock file into is acceptable.

#### Revision diff summary

### Testing
Changes have been manually tested, wrote some basic unit tests to go along with the changes.


RogerZhongAWS (commented):
I should probably move the call to ProcessLock() to somewhere after the CliArgs have been parsed. That way it does not interfere with users using the --help option

marcoemorais-aws (changes requested):

source/main.cpp
```diff
@@ -270,6 +272,7 @@ namespace Aws
 
 int main(int argc, char *argv[])
 {
+    LockFileUtils::ProcessLock();
```
HarshGandhi-AWS, line 275: we should lock file after MQTT connection is established since `abort()` is called if MQTT connection fails to establish.

RogerZhongAWS:
I think we still need to write the pid into the lockfile to handle cases where the program terminates unexpectedly, whether it is from abort() or some other crash. There are bits of code in my first implementation that I can use within the context of this new implementation.

source/main.cpp
RogerZhongAWS: I am not sure if this will lead to desired behavior. Establishing a new MQTT connection with an existing one already running will cause both connections to disconnect and attempt reconnection, which is the original problem acquiring the lock is trying to prevent.

RogerZhongAWS:
Removed calls to flockfile() since it only applies to threads. Now using lockf(), which is a wrapper around the fcntl lock call, which is bound to the process itself. Currently not sure if a lock is needed at all and if so, what would be the right type of file lock to use and why.

marcoemorais-aws (changes requested):

source/util/LockFile.cpp
```diff
@@ -0,0 +1,63 @@
...
+        string storedPid;
+        if (fileIn >> storedPid && !(kill(stoi(storedPid), 0) == -1 && errno == ESRCH))
+        {
+            string processPath = "/proc/" + storedPid + "/cmdline";
+            string cmdline;
+            ifstream cmd(processPath.c_str());
+            // check if process contains name
+            if (cmd && cmd >> cmdline && cmdline.find(process) != string::npos)
+            {
+                LOGM_ERROR(TAG, "Pid associated with active process found in lockfile: %s", filename.c_str());
```
marcoemorais-aws, line 31: Include the process name in the log message eg `Pid associated with active process named %s in lockfile: %s"`

source/util/LockFile.cpp
```diff
@@ -0,0 +1,59 @@
...
+    // remove stale pid file
+    remove(filename.c_str());
+    fileIn.close();
+
+    FILE *file = fopen(filename.c_str(), "wx");
+    if (!file || lockf(fileno(file), F_LOCK, 0))
+    {
+        LOGM_ERROR(TAG, "Unable to open lockfile", filename.c_str());
+
+        throw runtime_error{"Can not write to lockfile."};
```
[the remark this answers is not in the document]
marcoemorais-aws: Actually I am wrong, apologize I wasn't reading closely enough and when `file == NULL`, then you don't need to call `fclose`.

marcoemorais-aws (dismissed):

test/util/TestLockFile.cpp
```diff
@@ -0,0 +1,103 @@
...
+TEST(LockFile, multipleFiles)
+{
+    char buf[FILENAME_MAX];
+    string path = getcwd(buf, FILENAME_MAX);
+    path += "/devicecl.lock";
+    unique_ptr<LockFile> lockFile = unique_ptr<LockFile>(new LockFile{path, "test-aws-iot-device-client"});
+
+    try
+    {
+        unique_ptr<LockFile> lockFile2 = unique_ptr<LockFile>(new LockFile{path, "test-aws-iot-device-client"});
```
marcoemorais-aws, line 52: Better to use `ASSERT_THROW` (applies to additional tests below) eg

```
EXPECT_THROW(unique_ptr<LockFile>(new LockFile{path, "test-aws-iot-device-client"}), std::runtime_error);
```

test/util/TestLockFile.cpp
```diff
@@ -0,0 +1,103 @@
...
+TEST(LockFile, staleFile)
+{
+    char buf[FILENAME_MAX];
+    string path = getcwd(buf, FILENAME_MAX);
+    path += "/devicecl.lock";
+
+    ofstream fileOut(path);
+    if (fileOut)
+    {
+        fileOut << "1234";
```
marcoemorais-aws, line 89: Pids are monotonically increasing, will recycle, and 1024 is not very low. To reduce probability of collison might be better to read `/proc/sys/kernel/pid_max` and use that for the stale pid.

marcoemorais-aws (approved):
👍  Thank you for continuously addressing the comments.

source/config/Config.cpp
```diff
@@ -252,6 +252,13 @@ bool PlainConfig::LoadFromEnvironment()
         }
     }
 
+    const char *lockFilePathStr = std::getenv("LOCK_FILE_PATH");
+    if (lockFilePathStr)
+    {
+        LOGM_DEBUG(Config::TAG, "Set LOCK_FILE_PATH=%s", lockFilePathStr);
```
HarshGandhi-AWS, line 258: Please validate the path ad save the `Expanded path` over here instead of the value received from env variable. Also `Sanitize` the expanded path passed in LOG statement.

source/config/Config.h
```diff
@@ -93,6 +93,7 @@ namespace Aws
                 Aws::Crt::Optional<std::string> thingName;
 
                 aws_mem_trace_level memTraceLevel{AWS_MEMTRACE_NONE};
+                std::string lockFilePath{"/run/lock/devicecl.lock"};
```
HarshGandhi-AWS, line 96: store this value in DEFAULT_LOCK_FILE_PATH constant. 

```
static constexpr char DEFAULT_LOCK_FILE_PATH[] = "/run/lock/devicecl.lock";
std::string lockFilePath{DEFAULT_LOCK_FILE_PATH};
```

source/main.cpp
```diff
@@ -294,6 +324,11 @@ int main(int argc, char *argv[])
         LOG_WARN(TAG, "Unable to append current working directory to PATH environment variable.");
     }
 
+    if (!init(argc, argv))
```
HarshGandhi-AWS, line 327: Add a comment or update the function name so that it is easy to understand the meaning of the following code statements.

docs/ENV.md
```diff
@@ -17,7 +17,13 @@ In addition to the application configuration settings described in [Configuring
     * The device client will also print the contents of the trace during shutdown when memory tracing is enabled.
     * When there are no pending allocations or memory trace is not enabled, then nothing is printed to the SDK log file.
     * Enabling memory allocation tracing has a nontrivial cost and we do not recommend that customers enable this by default for production deployments.
-
+    
+* `LOCK_FILE_PATH`
+  * To enforce single instance creation, set `LOCK_FILE_PATH` to a file in a writeable directory. For example, if you want to write to a file named devicecl.lock in /run/lock, then `LOCK_FILE_PATH=/run/lock/devicecl.lock`. Permissions still apply when writing to restricted directories.
+  * This feature is disabled until the environment variable is set.
```
HarshGandhi-AWS, line 23: How is this true if we have a default value set for `LOCK_FILE_PATH`?

docs/ENV.md
```diff
@@ -17,7 +17,13 @@ In addition to the application configuration settings described in [Configuring
     * The device client will also print the contents of the trace during shutdown when memory tracing is enabled.
     * When there are no pending allocations or memory trace is not enabled, then nothing is printed to the SDK log file.
     * Enabling memory allocation tracing has a nontrivial cost and we do not recommend that customers enable this by default for production deployments.
-
+    
+* `LOCK_FILE_PATH`
+  * To enforce single instance creation, set `LOCK_FILE_PATH` to a file in a writeable directory. For example, if you want to write to a file named devicecl.lock in /run/lock, then `LOCK_FILE_PATH=/run/lock/devicecl.lock`. Permissions still apply when writing to restricted directories.
```
HarshGandhi-AWS, line 22: you can also consider this file location for lock file.

```
LOCK_FILE_PATH=~~/.aws-iot-device-client/devicecl.lock
```

docs/ENV.md
RogerZhongAWS: forgot to update the documentation, sorry

source/util/LockFile.h
```diff
@@ -0,0 +1,36 @@
...
+            namespace Util
+            {
+                struct LockFile
+                {
+                  private:
+                    static constexpr char TAG[] = "LockFile.cpp";
+                    std::string filename;
+
+                  public:
+                    explicit LockFile(const std::string &filename, const std::string &process);
```
HarshGandhi-AWS, line 24: Can you add some documentation over here regarding this function and its use? You can check `StringUtils.h` for reference.

RogerZhongAWS: sure

test/util/TestLockFile.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+TEST(LockFile, staleFile)
+{
+    string path = "/run/lock/";
+    string pidMax;
+    ifstream pidFile("/proc/sys/kernel/pid_max");
+    if (pidFile && pidFile >> pidMax)
+    {
+        ofstream fileOut(path);
```
marcoemorais-aws, line 62: [bug] This line will open `/run/lock` for output. Since this is a directory, the call will fail and nothing is written. I think your intent was to create a file named `path + "devicecl.lock"`. Keep it DRY by creating a variable named `filename` and using that in both places in this branch.

marcoemorais-aws (changes requested):

test/util/TestLockFile.cpp
RogerZhongAWS: Actually, I changed the code to no longer require the full path including the filename.

You now only have to specify the parent directory and device client will just create a file named "devicecl.lock" in that dir.

I suppose this takes away the option for giving the lockfile a different name but I don't know if such a use case will ever come up.

marcoemorais-aws (dismissed):

HarshGandhi-AWS (approved):

marcoemorais-aws (approved):

HarshGandhi-AWS (approved):

PR #242, opened 2022-04-14
Fix logging race condition

RogerZhongAWS:
### Motivation
- One of our unit tests has been failing due to a race condition bug in our logging code.

#### Change summary
replaced wait() with wait_for(), which establishes a maximum duration for the thread to unblock itself if the condition variable never gets notified.

### Testing
- tested via code tracing and lots of test runs, will need additional time to completely verify fix.

### DIFF
--- a/.github/build.sh
+++ b/.github/build.sh
@@ -233,3 +233,4 @@ cmake --build . --target test-aws-iot-device-client
 
 ### Run Tests ###
 env AWS_CRT_MEMORY_TRACING=1 ./test/test-aws-iot-device-client
+./test/test-aws-iot-device-client --gtest_filter=LogQueueTest.notifyAllOnShutdown --gtest_repeat=1000
--- a/source/logging/LogQueue.cpp
+++ b/source/logging/LogQueue.cpp
@@ -25,8 +25,7 @@ std::unique_ptr<LogMessage> LogQueue::getNextLog()
     unique_lock<mutex> readLock(queueLock);
     while (logQueue.empty() && !isShutdown)
     {
-        newLogNotifier.wait(readLock);
-        // newLogNotifier.wait_for(readLock, chrono::milliseconds (100));
+        newLogNotifier.wait_for(readLock, chrono::milliseconds (100));
     }
     if (logQueue.empty())
     {
--- a/source/logging/LogQueue.cpp
+++ b/source/logging/LogQueue.cpp
@@ -25,7 +25,7 @@ std::unique_ptr<LogMessage> LogQueue::getNextLog()
     unique_lock<mutex> readLock(queueLock);
     while (logQueue.empty() && !isShutdown)
     {
-        newLogNotifier.wait_for(readLock, chrono::milliseconds (100));
+        newLogNotifier.wait_for(readLock, chrono::milliseconds(100));
     }
     if (logQueue.empty())
     {
--- a/source/logging/LogQueue.cpp
+++ b/source/logging/LogQueue.cpp
@@ -26,7 +26,7 @@ std::unique_ptr<LogMessage> LogQueue::getNextLog()
     while (logQueue.empty() && !isShutdown)
     {
         newLogNotifier.wait(readLock);
-        //newLogNotifier.wait_for(readLock, chrono::milliseconds (100));
+        // newLogNotifier.wait_for(readLock, chrono::milliseconds (100));
     }
     if (logQueue.empty())
     {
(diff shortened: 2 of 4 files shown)


RogerZhongAWS:
the only question is how long to set the rtime argument for

source/logging/LogQueue.cpp
```diff
@@ -25,7 +25,7 @@ std::unique_ptr<LogMessage> LogQueue::getNextLog()
     unique_lock<mutex> readLock(queueLock);
     while (logQueue.empty() && !isShutdown)
     {
-        newLogNotifier.wait(readLock);
+        newLogNotifier.wait_for(readLock, chrono::seconds(1));
```
marcoemorais-aws, line 28: In the most pessimistic scenario (no log messages in queue), this setting will affect how soon after a shutdown the logger process will exit (FileLogger run loop pasted below).  As a result I think the timeout should be relatively short, perhaps 100ms eg `chrono::milliseconds(100)`.

```
void FileLogger::run()
{
    unique_lock<mutex> runLock(isRunningLock);
    isRunning = true;
    runLock.unlock();

    while (!needsShutdown)
    {
        unique_ptr<LogMessage> message = logQueue->getNextLog();

        if (NULL != message)
        {
            writeLogMessage(std::move(message));
        }
        this_thread::sleep_for(std::chrono::milliseconds(DEFAULT_WAIT_TIME_MILLISECONDS));
    }
}
```

marcoemorais-aws:
@RogerZhongAWS I have not been able to reproduce the flakey test on my local machine, although I have observed it on ci. Can you try adding an additional step to the CI build to repeat the test (as shown below) with and without this change so that we can have confidence that your change resolves the issue?

```
./test/test-aws-iot-device-client --gtest_filter=LogQueueTest.notifyAllOnShutdown --gtest_repeat=100
```

source/logging/LogQueue.cpp
HarshGandhi-AWS, line 28: I think this change will increase the CPU consumption since the while condition is validated after `x` seconds in loop.

RogerZhongAWS: You are right, this change will certainly cost additional CPU cycles every x seconds. The current fix is the simplest change we can make right now, which can be fine if our overall CPU consumption is still low. However, if this is too big of an issue, then we will have to figure out another way to fix the bug.

HarshGandhi-AWS: In that case lets test everything (CPU consumption) before pushing this change in mainline.

RogerZhongAWS: Setting the timeout period to 1 millisecond consumes 1% CPU on my ec2 instance. Anything over 100 milliseconds has
Screen Shot 2022-04-18 at 5 04 59 PM
 0.0 - 0.1% CPU usage. Of course we cannot assume anything about customer cpu performance but I think something like 200ms should be safe for now

RogerZhongAWS: Perhaps in the future we can optimize the code further to achieve even less resource consumption but the logger would need to be reworked to ensure correctness at the same time.

test/logging/TestLogQueue.cpp
```diff
@@ -85,8 +85,8 @@ TEST_F(LogQueueTest, notifyAllOnShutdown)
 
     logQueue->shutdown();
 
-    cv.wait_for(lock1, chrono::seconds(1));
-    cv2.wait_for(lock2, chrono::seconds(1));
+    cv.wait_for(lock1, chrono::milliseconds(400));
```
marcoemorais-aws, line 88: [Minor] I guess you can make it `2 * EMPTY_WAIT_TIME_MILLISECONDS` to be a little more explicit about the connection between this wait and the code that it relies on, but I will leave it to you to decide.

marcoemorais-aws (dismissed):

test/logging/TestLogQueue.cpp
RogerZhongAWS: since EMPTY_WAIT_TIME_MILLISECONDS is a private member, I'll leave it as is and add a comment for explanation

marcoemorais-aws (approved):

.github/build.sh
```diff
@@ -233,3 +233,4 @@ cmake --build . --target test-aws-iot-device-client
 
 ### Run Tests ###
 env AWS_CRT_MEMORY_TRACING=1 ./test/test-aws-iot-device-client
+./test/test-aws-iot-device-client --gtest_filter=LogQueueTest.notifyAllOnShutdown --gtest_repeat=1000
```
marcoemorais-aws, line 236: Keep this in for now, but comment this out in the future after we have confidence that the issue is resolved.

HarshGandhi-AWS (approved):

PR #243, opened 2022-04-20
implemented max logging size

RogerZhongAWS:
#### Change summary
Caps max logging message size to a certain value. However, we do log the entire config file on startup so that message will be truncated if the max buffer size is too small

Documentation about null terminated strings has already been done at an earlier time.

### DIFF
--- a/source/util/StringUtils.cpp
+++ b/source/util/StringUtils.cpp
@@ -2,9 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "StringUtils.h"
+#include "../config/Config.h"
 #include <cstdarg>
 #include <map>
-#include <sstream>
 
 using namespace std;
 
@@ -20,8 +20,12 @@ namespace Aws
                 {
                     va_list copy;
                     va_copy(copy, args);
-                    int formattedMsgSize = vsnprintf(nullptr, 0, message, copy) + 1;
+                    size_t formattedMsgSize = vsnprintf(nullptr, 0, message, copy);
                     va_end(copy);
+
+                    // will format up to MAX_CONFIG_SIZE, the maximum buffer size after which log messages will get
+                    // truncated
+                    formattedMsgSize = std::min(formattedMsgSize, Config::MAX_CONFIG_SIZE - 1) + 1;
 
                     std::string buffer;
                     buffer.resize(formattedMsgSize);
--- a/source/config/Config.cpp
+++ b/source/config/Config.cpp
@@ -1998,14 +1998,14 @@ bool Config::ParseConfigFile(const string &file, bool isRuntimeConfig)
     }
 
     size_t incomingFileSize = FileUtils::GetFileSize(file);
-    if (5000 < incomingFileSize)
+    if (incomingFileSize > Config::MAX_CONFIG_SIZE)
     {
         LOGM_WARN(
             TAG,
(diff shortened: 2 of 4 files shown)


marcoemorais-aws (changes requested):

marcoemorais-aws (approved):

marcoemorais-aws:
Might be good to add a pair of tests that verify messages with and without truncation.

source/util/StringUtils.cpp
```diff
@@ -16,13 +16,17 @@ namespace Aws
         {
             namespace Util
             {
+                // maximum buffer size after which log messages will get truncated
+                static const int MAX_BUF_SIZE = 4096;
```
HarshGandhi-AWS, line 20: nit: in future you can use unsigned `int8_t` or `short`

source/util/StringUtils.cpp
```diff
@@ -16,13 +16,17 @@ namespace Aws
...
                 string vFormatMessage(const char *message, va_list args)
                 {
                     va_list copy;
                     va_copy(copy, args);
-                    int formattedMsgSize = vsnprintf(nullptr, 0, message, copy) + 1;
+                    int formattedMsgSize = vsnprintf(nullptr, 0, message, copy);
                     va_end(copy);
 
+                    formattedMsgSize = std::min(formattedMsgSize, MAX_BUF_SIZE - 1) + 1;
```
HarshGandhi-AWS, line 28: Since the log message bigger than `4096` will be dropped, we should document it.

RogerZhongAWS: Do you think this is a reasonable size? Upon device client startup we log the entire config file, which is why I increased the limit. It is a little bit of an arbitrary decision.

marcoemorais-aws: We cap the config file at 5k. Could align this limit to that constant which is just a magic number (linked below).
https://github.com/awslabs/aws-iot-device-client/blob/140913a61b35f1fed53506bf94fe96425f79f33c/source/config/Config.cpp#L1616

RogerZhongAWS: yes

marcoemorais-aws: @RogerZhongAWS  Can you make this a public constant visible and shared by Config.cpp and StringUtils.cpp?

marcoemorais-aws: @RogerZhongAWS  Can you make this a public constant visible and shared by Config.cpp and StringUtils.cpp?

source/util/StringUtils.cpp
```diff
@@ -17,7 +17,7 @@ namespace Aws
             namespace Util
             {
                 // maximum buffer size after which log messages will get truncated
-                static const int MAX_BUF_SIZE = 4096;
+                static const u_int16_t MAX_BUF_SIZE = 5000;
```
marcoemorais-aws, line 20: `u_int16_t` is windows-specific, use `std::uint16_t` (from `<cstdint>` header) for cross-platform code. For this use case I recommend using `std::size_t` (from `<cstddef>`)  to match the type of `vsnprintf`.

https://en.cppreference.com/w/cpp/header/cstdint

marcoemorais-aws (dismissed):

marcoemorais-aws (dismissed):

marcoemorais-aws (approved):

HarshGandhi-AWS (approved):

PR #247, opened 2022-04-28
Root ca

shangabl:
### Motivation
The SDK changed such that a bad root-ca path would override the trust store. This change makes root-ca optional so that the SDK will default to trust store.

#### Change summary
Makes root-ca optional and adds a null check to ST connect method to prevent seg fault.

### Testing
Manually tested by opening and closing tunnels

### DIFF
--- a/source/tunneling/SecureTunnelingContext.cpp
+++ b/source/tunneling/SecureTunnelingContext.cpp
@@ -36,10 +36,7 @@ namespace Aws
                     mOnConnectionShutdown = onConnectionShutdown;
                 }
 
-                SecureTunnelingContext::~SecureTunnelingContext()
-                {
-                    mSecureTunnel->Close();
-                }
+                SecureTunnelingContext::~SecureTunnelingContext() { mSecureTunnel->Close(); }
 
                 template <typename T>
                 static bool operator==(const Aws::Crt::Optional<T> &lhs, const Aws::Crt::Optional<T> &rhs)
@@ -110,7 +107,8 @@ namespace Aws
                         bind(&SecureTunnelingContext::OnStreamReset, this),
                         bind(&SecureTunnelingContext::OnSessionReset, this)));
 
-                    bool success = mSecureTunnel->GetUnderlyingHandle() != nullptr && mSecureTunnel->Connect() == AWS_OP_SUCCESS;
+                    bool success =
+                        mSecureTunnel->GetUnderlyingHandle() != nullptr && mSecureTunnel->Connect() == AWS_OP_SUCCESS;
                     if (!success)
                     {
                         LOG_ERROR(TAG, "Cannot connect to secure tunnel. Please see the SDK log for detail.");
--- a/source/tunneling/SecureTunnelingContext.cpp
+++ b/source/tunneling/SecureTunnelingContext.cpp
@@ -38,10 +38,7 @@ namespace Aws
 
                 SecureTunnelingContext::~SecureTunnelingContext()
                 {
-                    if(mSecureTunnel->GetUnderlyingHandle() != nullptr)
-                    {
-                        mSecureTunnel->Close();
-                    }
+                    mSecureTunnel->Close();
                 }
 
                 template <typename T>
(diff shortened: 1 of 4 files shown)


HarshGandhi-AWS (commented):
RootCA is still a [required parameter](https://github.com/awslabs/aws-iot-device-client/blob/868465b5c8faebd55a727c1629806d93b493bb9a/source/config/Config.cpp#L287-L290) in config parser. Please update that.

source/tunneling/SecureTunnelingFeature.cpp
```diff
@@ -102,7 +102,7 @@ namespace Aws
                         std::unique_ptr<SecureTunnelingContext> context =
                             unique_ptr<SecureTunnelingContext>(new SecureTunnelingContext(
                                 mSharedCrtResourceManager,
-                                *config.rootCa,
+                                mRootCa,
                                 *config.tunneling.destinationAccessToken,
                                 GetEndpoint(*config.tunneling.region),
                                 static_cast<uint16_t>(config.tunneling.port.value()),
```
RogerZhongAWS, line 108: nit: this isn't too relevant to your PR, but I wonder why sometimes we use the * operator to get the optional value and other times we call value(). Looking into the Aws::Crt::Optional type, it seems like the behavior is the same for both methods of access.

source/tunneling/SecureTunnelingContext.cpp
```diff
@@ -22,14 +22,14 @@ namespace Aws
...
-                    const string &rootCa,
+                    Aws::Crt::Optional<std::string> &rootCa,
                     const string &accessToken,
                     const string &endpoint,
                     uint16_t port,
                     OnConnectionShutdownFn onConnectionShutdown)
                 {
                     mSharedCrtResourceManager = manager;
-                    mRootCa = rootCa;
+                    mRootCa = rootCa.has_value() ? rootCa.value() : basic_string<char>("");
```
RogerZhongAWS, line 32: Just want to highlight this part to go along with my other comment.

source/tunneling/SecureTunnelingContext.cpp
```diff
@@ -101,12 +107,13 @@ namespace Aws
                         bind(&SecureTunnelingContext::OnStreamReset, this),
                         bind(&SecureTunnelingContext::OnSessionReset, this)));
 
-                    bool ok = mSecureTunnel->Connect() == AWS_OP_SUCCESS;
-                    if (!ok)
+                    bool success =
```
BolongZhang-AWS, line 110: nit: usually we name these like `connnectSuccess` for better understanding

source/tunneling/SecureTunnelingContext.cpp
```diff
@@ -22,14 +22,14 @@ namespace Aws
...
                 SecureTunnelingContext::SecureTunnelingContext(
                     shared_ptr<SharedCrtResourceManager> manager,
-                    const string &rootCa,
+                    Aws::Crt::Optional<std::string> &rootCa,
                     const string &accessToken,
                     const string &endpoint,
                     uint16_t port,
                     OnConnectionShutdownFn onConnectionShutdown)
                 {
                     mSharedCrtResourceManager = manager;
```
marcoemorais-aws, line 31: Two recommendations.
* Make it a const reference eg `const Aws:...`
* Replace assignment in constructor body with use of member initialization list eg `: mSharedCrtResourceManager(manager), mRootCa(rootCa.has_value() ? rootCa.value() : basic_string<char>("")), ...`

marcoemorais-aws (changes requested):

source/tunneling/SecureTunnelingFeature.cpp
marcoemorais-aws: Technically equivalent, but sometimes operator precedence rules require wrapping the dereferenced value in parenthesis eg `(*optional_thing)->` This would be most common when the optional thing you have is actually a pointer to something else and you want to use the member access operator `->` on the dereferenced thing.  You can see from the precedence table below that member access is higher precedence than dereference hence you need the parens. In contrast, use of `value()` doesn't require the extra parenthesis.

https://en.cppreference.com/w/cpp/language/operator_precedence

marcoemorais-aws (approved):

RogerZhongAWS (approved):

PR #248, opened 2022-04-29
Enable valgrind as part of build

marcoemorais-aws:
### Motivation
* Add memory leak detection when running unit tests using valgrind.

### Modifications
* Add valgrind to the base image.
* Run valgrind when running unit tests from build.sh.
* Add custom target named `test-aws-iot-device-client-leak-check` to cmake that runs the unit tests with valgrind in the same way as build.sh.

### Testing
* Unit tests only.

### DIFF
--- a/.github/build.sh
+++ b/.github/build.sh
@@ -232,4 +232,4 @@ fi
 cmake --build . --target test-aws-iot-device-client
 
 ### Run Tests ###
-env AWS_CRT_MEMORY_TRACING=1 valgrind --leak-check=yes ./test/test-aws-iot-device-client
+env AWS_CRT_MEMORY_TRACING=1 valgrind --leak-check=yes --error-exitcode=1 ./test/test-aws-iot-device-client
--- a/.github/build.sh
+++ b/.github/build.sh
@@ -232,5 +232,4 @@ fi
 cmake --build . --target test-aws-iot-device-client
 
 ### Run Tests ###
-env AWS_CRT_MEMORY_TRACING=1 ./test/test-aws-iot-device-client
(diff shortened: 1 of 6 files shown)


.github/build.sh
```diff
@@ -232,5 +232,4 @@ fi
 cmake --build . --target test-aws-iot-device-client
 
 ### Run Tests ###
-env AWS_CRT_MEMORY_TRACING=1 ./test/test-aws-iot-device-client
-./test/test-aws-iot-device-client --gtest_filter=LogQueueTest.notifyAllOnShutdown --gtest_repeat=1000
```
marcoemorais-aws, line 236: @RogerZhongAWS The notify unit test has been repeating for 1k times over the past week and we haven't seen any flaky tests.  I think it is safe to remove the repeat. Please let me know if you disagree.

RogerZhongAWS: Agreed

.github/docker-images/ubi8/Dockerfile
```diff
@@ -36,7 +37,7 @@ RUN curl -sSL https://github.com/Kitware/CMake/releases/download/v3.10.0/cmake-3
     && tar -zxvf cmake-3.10.0.tar.gz \
     && cd cmake-3.10.0 \
     && ./bootstrap \
-    && make -j 4 \
+    && make \
```
HarshGandhi-AWS, line 40: QQ: why only in this docker file this change is made?

HarshGandhi-AWS (approved):

RogerZhongAWS (approved):

.github/docker-images/ubi8/Dockerfile
marcoemorais-aws: It actually doesn't make sense to explicitly set parallelism here since the amount of parallelism depends on how many cores allocated to docker build.

Issue #146, opened 2021-06-08
Build failure with gcc v11.*


HarshGandhi-AWS:
**Describe the bug**

Failing to build Device Client while using GCC 11. 
ERROR LOGS: 
```
 25%] Building C object aws-iot-device-sdk-cpp-v2-build/crt/aws-crt-cpp/crt/s2n/CMakeFiles/s2n.dir/pq-crypto/kyber_90s_r2/ntt.c.o
/home/ubuntu/1.2/aws-iot-device-client/build/aws-iot-device-sdk-cpp-v2-src/crt/aws-crt-cpp/crt/s2n/pq-crypto/kyber_90s_r2/ntt.c:92:44: error: argument 1 of type ‘int16_t[256]’ {aka ‘short int[256]’} with mismatched bound [-Werror=array-parameter=]
   92 | void PQCLEAN_KYBER51290S_CLEAN_ntt(int16_t poly[256]) {
      |                                    ~~~~~~~~^~~~~~~~~
.
.
.
```

**To Reproduce**

1. Install GCC v11
2. Build Device client by following the Quick Start guide from README. 

**Expected behavior**

Successfully build Device Client binary

**Actual behavior**

Build failure with error on aws-iot-device-sdk-cpp-v2 side. (Error mentioned above)

**Full Error Logs**

```
[ 24%] Building C object aws-iot-device-sdk-cpp-v2-build/crt/aws-crt-cpp/crt/s2n/CMakeFiles/s2n.dir/pq-crypto/kyber_90s_r2/kyber_90s_r2_kem.c.o
[ 25%] Building C object aws-iot-device-sdk-cpp-v2-build/crt/aws-crt-cpp/crt/s2n/CMakeFiles/s2n.dir/pq-crypto/kyber_90s_r2/ntt.c.o
/home/ubuntu/1.2/aws-iot-device-client/build/aws-iot-device-sdk-cpp-v2-src/crt/aws-crt-cpp/crt/s2n/pq-crypto/kyber_90s_r2/ntt.c:92:44: error: argument 1 of type ‘int16_t[256]’ {aka ‘short int[256]’} with mismatched bound [-Werror=array-parameter=]
   92 | void PQCLEAN_KYBER51290S_CLEAN_ntt(int16_t poly[256]) {
      |                                    ~~~~~~~~^~~~~~~~~
[...]

PR #249, opened 2022-05-03
bump ubuntu version, fix build issues with gcc 11

RogerZhongAWS:
### Motivation
- Build with gcc 11 fails when calling this_thread::sleep_for.
- Found when investigating the issue below.
- Issue number: #146 

#### Change summary
Added missing headers to fix.

### DIFF
--- a/.github/docker-images/ubuntu-16-x64/Dockerfile
+++ b/.github/docker-images/ubuntu-16-x64/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 ENV DEBIAN_FRONTEND=noninteractive
 
@@ -8,13 +8,10 @@ ARG OPENSSL_VERSION=1.1.1n
 # Install prereqs
 ###############################################################################
 RUN apt-get update -qq \
-    && add-apt-repository ppa:ubuntu-toolchain-r/test \
     && apt-get -y install \
     git \
     curl \
     build-essential \
-    gcc-11 \
-    g++-11 \
     wget \
     cppcheck \
     libc6-dbg \
@@ -47,7 +44,7 @@ RUN curl -sSL https://github.com/Kitware/CMake/releases/download/v3.10.0/cmake-3
 # Clone and build Google Test
 ###############################################################################
 WORKDIR /tmp
-RUN wget https://github.com/google/googletest/archive/release-1.10.0.tar.gz \
+RUN wget https://github.com/google/googletest/archive/release-1.10.0.tar.gz --no-check-certificate \
     && tar xf release-1.10.0.tar.gz \
     && cd googletest-release-1.10.0 \
     && cmake -DBUILD_SHARED_LIBS=ON . \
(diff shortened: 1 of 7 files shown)


marcoemorais-aws (dismissed):

marcoemorais-aws:
@RogerZhongAWS One suggestion, can you add gcc-11 to the testing matrix in ci.yml? 

Should be as simple as changing from `version: [ 5,6,7,8 ]` to `version: [ 5,6,7,8,11 ]`.

https://github.com/awslabs/aws-iot-device-client/blob/5694b5ac14a4622deeb0316e957459ce551ad82e/.github/workflows/ci.yml#L122

philcali:
FWIW: I just ran into this issue, and confirmed this PR fixes it for me (ubuntu:latest x86_64).

.github/docker-images/ubuntu-16-x64/Dockerfile
```diff
@@ -44,7 +44,7 @@ RUN curl -sSL https://github.com/Kitware/CMake/releases/download/v3.10.0/cmake-3
 # Clone and build Google Test
 ###############################################################################
 WORKDIR /tmp
-RUN wget https://github.com/google/googletest/archive/release-1.10.0.tar.gz \
+RUN wget https://github.com/google/googletest/archive/release-1.10.0.tar.gz --no-check-certificate \
```
marcoemorais-aws, line 47: Ignoring the certificate chain opens us up to man in the middle (MITM) attacks on the software supply chain (eg dependencies used to build the software).  We should never do this. The likely reason for adding this flag is that wget is not finding the certificate store in the operating system. The solution is to install the package containing the certificates. Please add the line below `after` installing prereqs but `before` downloading dependencies.

```
RUN apt-get install -y ca-certificates
```

.github/workflows/ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Build ${{ env.PACKAGE_NAME }}
         run: |
           echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u awslabs --password-stdin
-          export DOCKER_IMAGE=ghcr.io/awslabs/aws-iot-device-client/ubuntu-16-x64:latest
+          export DOCKER_IMAGE=ghcr.io/awslabs/aws-iot-device-client/ubuntu-18-x64:latest
```
marcoemorais-aws, line 53: * You should rename the label (line 43) from `build-ubuntu-16-x64` to `build-ubuntu-18-x64`.  
* Also I think we have to replace all instances in this file `s/ubuntu-16-x64/ubuntu-18-x64/g` since we are essentially bumping our ubuntu image from 16 to 18.

marcoemorais-aws (changes requested):

.github/docker-images/ubuntu-16-x64/Dockerfile
RogerZhongAWS: Thanks for the heads up. I remember I did try doing that install but it still had issues verifying the cert. I haven't tested to see if the github workflow is able to complete yet, but hopefully I can find a fix for this asap.

marcoemorais-aws (approved):

HarshGandhi-AWS (approved):

PR #258, opened 2022-05-04
Root ca fix

shangabl:
### Motivation
The GG ST Component passes a placeholder value for root-ca which was causing failures.

#### Change summary
Use an empty string when root-ca path is invalid.

### Testing
Manually tested by opening tunnels.

### DIFF
--- a/source/config/Config.cpp
+++ b/source/config/Config.cpp
@@ -113,7 +113,16 @@ bool PlainConfig::LoadFromJson(const Crt::JsonView &json)
     {
         if (!json.GetString(jsonKey).empty())
         {
-            rootCa = FileUtils::ExtractExpandedPath(json.GetString(jsonKey).c_str());
+            basic_string<char> path = FileUtils::ExtractExpandedPath(json.GetString(jsonKey).c_str());
+            if (!FileUtils::FileExists(path))
+            {
+                LOGM_ERROR(Config::TAG, "Path %s to RootCA is invalid. Ignoring... Will attempt to use default trust store.", path.c_str());
+                rootCa = "";
+            }
+            else
+            {
+                rootCa = FileUtils::ExtractExpandedPath(json.GetString(jsonKey).c_str());
+            }
         }
         else
         {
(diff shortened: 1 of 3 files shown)


source/config/Config.cpp
```diff
@@ -113,7 +113,19 @@ bool PlainConfig::LoadFromJson(const Crt::JsonView &json)
     {
         if (!json.GetString(jsonKey).empty())
         {
-            rootCa = FileUtils::ExtractExpandedPath(json.GetString(jsonKey).c_str());
+            basic_string<char> path = FileUtils::ExtractExpandedPath(json.GetString(jsonKey).c_str());
```
marcoemorais-aws, line 116: Can you add a test for this as well?

marcoemorais-aws (changes requested):

source/config/Config.cpp
```diff
@@ -113,7 +113,19 @@ bool PlainConfig::LoadFromJson(const Crt::JsonView &json)
...
+            if (!FileUtils::FileExists(path))
+            {
+                LOGM_ERROR(
```
marcoemorais-aws, line 119: If we aren't going to fail device client due to an invalid path, then level should be `LOGM_WARN` instead of `LOGM_ERROR`.

marcoemorais-aws (changes requested):

source/config/Config.cpp
marcoemorais-aws: You also need similar logic in `PlainConfig::LoadFromCliArgs`. https://github.com/awslabs/aws-iot-device-client/blob/c53400022f87ac0cb7fe8b8f3bb18e4f5198ca26/source/config/Config.cpp#L225

source/config/Config.cpp
shangabl: I'm actually appending it to the return value of GetString to pass to ExtractExpandedPath.

test/config/TestConfig.cpp
```diff
@@ -21,6 +21,7 @@ using namespace Aws::Iot::DeviceClient;
 using namespace Aws::Iot::DeviceClient::Util;
 
 const string filePath = "/tmp/aws-iot-device-client-test-file";
+const string invalidFilePath = "/tmp/invalid-file-path";
```
marcoemorais-aws, line 24: Please add a `std::remove(invalidFilePath)` on the `ConfigTestFixture::SetUp` method to be sure.

marcoemorais-aws (changes requested):

marcoemorais-aws (approved):

HarshGandhi-AWS (approved):

Issue #173, opened 2021-10-14
systemd script does not have enough info to run


rpcme:
**Describe the bug**

The systemd script runs /sbin/aws-iot-device-client with no options.

No one would ever install configuration, certificate, key, and so
forth to home directory, which is the default, under this condition.

The script should have --config-file flag with the default system
volatile install location which is /etc. Please define the expected
location for this file in documentation, and then configure the proper
system config file in the systemd script.

**To Reproduce**

look at the systemd script here

https://github.com/awslabs/aws-iot-device-client/blob/main/setup/aws-iot-device-client.service

**Expected behavior**

systemd script should invoke aws-iot-device-client with --config-file
or have sane system level defaults for key configuration and secure
data.

**Actual behavior**

Please see bug section

**Logs**

Not applicable

**Environment (please complete the following information):**
 - OS: all
 - Version: 1.4.0
 - Architecture: x86-64, AARCH64, RISC-V
 - Device Client commit: tag v1.4.0

**Additional context**

none

satheshantanu:
Hi @rpcme thanks for this - we are working to build the solution to this issue.

PR #259, opened 2022-05-05
Improvement to device client setup

RogerZhongAWS:
### Motivation
- The systemd script should have --config-file flag with default directory in /etc
- Issue number: #173

#### Change summary
Change fixes the issue, updates documentation and added a small QOL improvement in the setup script.

### Testing
Manually tested ./setup.sh


setup.sh
```diff
@@ -282,17 +282,22 @@ if [ "$BUILD_CONFIG" = "y" ]; then
...
-    read -r GOOD_TO_GO
-    if [ "$GOOD_TO_GO" = "y" ]; then
-      CONFIGURED=1
-      mkdir -p "$OUTPUT_DIR"
-      echo "$CONFIG_OUTPUT" | tee "$CONF_OUTPUT_PATH" >/dev/null
-      chmod 745 "$OUTPUT_DIR"
-      chmod 644 "$CONF_OUTPUT_PATH"
-      printf ${GREEN} "Configuration has been successfully written to ${CONF_OUTPUT_PATH}"
-    fi
+    while [ "$CONFIRMED" != 1 ]; do
```
BolongZhang-AWS, line 285: If user enters "n" then it will just keep asking that question without letting them to edit/correct?

RogerZhongAWS: If they enter 'n' it should go back to the beginning

If they enter anything else than 'y' or 'n' it will ask the question again.

I'm mainly doing this because I have had an issue myself where I accidentally press enter and I have to enter everything over again.

RogerZhongAWS: I suppose experienced users of device client will just create a config file from a template but for those who are trying it out for the first time, maybe this will help them in case they press enter by accident

shangabl (approved):

BolongZhang-AWS (approved):

xlcheng1 (approved):

PR #261, opened 2022-05-07
Root ca

HarshGandhi-AWS:
### Motivation
- We should not validate code while parsing. We have a separate method `Validate` for this purpose. 
- Device Client should fail **only** if invalid RootCA file path is passed. Since it is not an mandatory field anymore, we should should not fail if the user fails to pass value for it. 

#### Change summary

* Removed redundant code and moved it to `Validate` method in config.cpp
* Updated the CONFIG.md documentation file. 

#### Revision diff summary

### Testing
- Added enough unit tests for testing my changes. 
- CI test run result: <link>


shangabl:
I made all these changes so that Greengrass Secure Tunneling component would work. I agree we should update the README but the rest of the changes are just undoing the changes that make fixed Greengrass.

shangabl (approved):

RogerZhongAWS (approved):

shangabl (approved):

PR #262, opened 2022-05-10
Root ca fix

shangabl:
### Motivation
Empty or invalid root-ca path would cause segmentation fault in the resource manager.

#### Change summary
This change will ignore empty or invalid root-ca paths. mqtt client will default to default trust store if no root-ca is passed.

### Testing
Manually tested with empty root-ca, invalid root-ca, then a valid root-ca path to an invalid root-ca.

### DIFF
--- a/source/SharedCrtResourceManager.cpp
+++ b/source/SharedCrtResourceManager.cpp
@@ -75,23 +75,6 @@ bool SharedCrtResourceManager::locateCredentials(const PlainConfig &config)
         }
     }
 
-    if (stat(config.rootCa->c_str(), &fileInfo) != 0)
-    {
-        LOGM_ERROR(
-            TAG, "Failed to find %s, cannot establish MQTT connection", Sanitize(config.rootCa->c_str()).c_str());
-        locatedAll = false;
-    }
-    else
-    {
-        string parentDir = FileUtils::ExtractParentDirectory(config.rootCa->c_str());
-        if (!FileUtils::ValidateFilePermissions(parentDir, Permissions::ROOT_CA_DIR) ||
-            !FileUtils::ValidateFilePermissions(config.rootCa->c_str(), Permissions::ROOT_CA))
-        {
-            LOG_ERROR(TAG, "Incorrect permissions on Root CA file and/or parent directory");
-            locatedAll = false;
-        }
-    }
-
     return locatedAll;
 }
 
@@ -252,7 +235,9 @@ int SharedCrtResourceManager::establishConnection(const PlainConfig &config)
     }
     auto clientConfigBuilder = MqttClientConnectionConfigBuilder(config.cert->c_str(), config.key->c_str());
     clientConfigBuilder.WithEndpoint(config.endpoint->c_str());
-    clientConfigBuilder.WithCertificateAuthority(config.rootCa->c_str());
+    if(config.rootCa.has_value()) {
+        clientConfigBuilder.WithCertificateAuthority(config.rootCa->c_str());
+    }
     clientConfigBuilder.WithSdkName(SharedCrtResourceManager::BINARY_NAME);
     clientConfigBuilder.WithSdkVersion(DEVICE_CLIENT_VERSION);
 
--- a/source/SharedCrtResourceManager.cpp
+++ b/source/SharedCrtResourceManager.cpp
@@ -235,7 +235,8 @@ int SharedCrtResourceManager::establishConnection(const PlainConfig &config)
     }
     auto clientConfigBuilder = MqttClientConnectionConfigBuilder(config.cert->c_str(), config.key->c_str());
     clientConfigBuilder.WithEndpoint(config.endpoint->c_str());
-    if(config.rootCa.has_value()) {
+    if (config.rootCa.has_value())
+    {
         clientConfigBuilder.WithCertificateAuthority(config.rootCa->c_str());
     }
     clientConfigBuilder.WithSdkName(SharedCrtResourceManager::BINARY_NAME);
--- a/source/SharedCrtResourceManager.cpp
+++ b/source/SharedCrtResourceManager.cpp
@@ -235,7 +235,7 @@ int SharedCrtResourceManager::establishConnection(const PlainConfig &config)
     }
     auto clientConfigBuilder = MqttClientConnectionConfigBuilder(config.cert->c_str(), config.key->c_str());
     clientConfigBuilder.WithEndpoint(config.endpoint->c_str());
-    if (config.rootCa.has_value())
+    if (config.rootCa.has_value() && !config.rootCa->empty())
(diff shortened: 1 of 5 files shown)


source/SharedCrtResourceManager.cpp
```diff
@@ -252,7 +235,10 @@ int SharedCrtResourceManager::establishConnection(const PlainConfig &config)
     }
     auto clientConfigBuilder = MqttClientConnectionConfigBuilder(config.cert->c_str(), config.key->c_str());
     clientConfigBuilder.WithEndpoint(config.endpoint->c_str());
-    clientConfigBuilder.WithCertificateAuthority(config.rootCa->c_str());
+    if (config.rootCa.has_value())
```
HarshGandhi-AWS, line 238: Also add `not empty` check.

source/SharedCrtResourceManager.h
```diff
@@ -44,14 +44,18 @@ namespace Aws
...
 
                 int buildClient(const PlainConfig &config);
 
                 void initializeAllocator(const PlainConfig &config);
 
+              protected:
+                /**
+                 * inheritable for testing
+                 */
+                bool locateCredentials(const PlainConfig &config);
```
HarshGandhi-AWS, line 57: If we are removing this method, then why are we not removing the code written in `source/SharedCrtResourceManager.cpp` file?

source/config/Config.cpp
```diff
@@ -121,6 +121,10 @@ bool PlainConfig::LoadFromJson(const Crt::JsonView &json)
             auto path = FileUtils::ExtractExpandedPath(json.GetString(jsonKey).c_str());
             if (FileUtils::FileExists(path))
             {
+                FileUtils::ValidateFilePermissions(
+                    FileUtils::ExtractParentDirectory(path), Permissions::ROOT_CA_DIR, false);
+                FileUtils::ValidateFilePermissions(path, Permissions::ROOT_CA, false);
+
                 rootCa = FileUtils::ExtractExpandedPath(json.GetString(jsonKey).c_str());
```
HarshGandhi-AWS, line 128: you can replace this line of code with: `rootCa = path` or something equivalent to avoid redundant code.

source/config/Config.cpp
```diff
@@ -121,6 +121,10 @@ bool PlainConfig::LoadFromJson(const Crt::JsonView &json)
             auto path = FileUtils::ExtractExpandedPath(json.GetString(jsonKey).c_str());
             if (FileUtils::FileExists(path))
             {
+                FileUtils::ValidateFilePermissions(
```
HarshGandhi-AWS, line 124: Why are we validating over here if we have the same code added in `Validate` method?

source/config/Config.cpp
```diff
@@ -121,6 +121,10 @@ bool PlainConfig::LoadFromJson(const Crt::JsonView &json)
...
+                    FileUtils::ExtractParentDirectory(path), Permissions::ROOT_CA_DIR, false);
+                FileUtils::ValidateFilePermissions(path, Permissions::ROOT_CA, false);
```
HarshGandhi-AWS, line 126: This should fail with fatal error. It is a security risk to allow files with incorrect file permissions.

HarshGandhi-AWS (commented):
Making RootCA optional is a core functionality change which will not benefit us at all but will only complicate things for us. I would still recommend making changes in GG ST component instead of making changes in DC for this.

source/SharedCrtResourceManager.h
shangabl: This is not being removed.

source/SharedCrtResourceManager.cpp
shangabl: I can add but rootCa will not have a value if it is passed as an empty string. Lots of tests to ensure that behavior.

RootCa is also optional now and this would fix the segmentation fault bug. The default behavior of the SDK is to use default trust store if no rootCa. What core functionality does this change?

test/config/TestConfig.cpp
```diff
@@ -235,18 +256,284 @@ TEST_F(ConfigTestFixture, InvalidRootCaPathConfig)
...
+    ASSERT_STREQ(rootCaPath.c_str(), config.rootCa->c_str());
+    ASSERT_STREQ("endpoint value", config.endpoint->c_str());
+    ASSERT_STREQ(filePath.c_str(), config.cert->c_str());
+    ASSERT_STREQ(filePath.c_str(), config.key->c_str());
+    ASSERT_STREQ("thing-name value", config.thingName->c_str());
+    ASSERT_FALSE(config.fleetProvisioning.enabled);
+    DefaultFeaturesEnabled(config);
+}
+
+#if defined(MQTT_DISABLED)
```
marcoemorais-aws, line 296: There is no macro named `MQTT_DISABLED` as a result these tests are never executed. You can confirm this by looking at the test output in CI for any of the build targets and observe that no test matching `ExplicitRootCaBadParentPermissionsConfig`. Before you spin another PR that replaces `MQTT_DISABLED` with `!defined(DISABLE_MQTT)` can you explain in the form of a comment why this definition becomes implicated in the test.  The reason I think this is important for exactly the kind of problem that you have stumbled onto -- preprocessor based includes can introduce hard to discover bugs.

marcoemorais-aws (changes requested):

test/config/TestConfig.cpp
```diff
@@ -612,8 +927,8 @@ TEST_F(ConfigTestFixture, DeviceDefenderCli)
 {
 	"endpoint": "endpoint value",
 	"cert": "/tmp/aws-iot-device-client-test-file",
+    "root-ca": "/tmp/aws-iot-device-client-test/AmazonRootCA1.pem",
```
marcoemorais-aws, line 930: Some minor formatting problem (same applies a few places below).

marcoemorais-aws:
Tests look good with exception of the preprocessor.

test/config/TestConfig.cpp
shangabl: Discussed already but for posterity:

When built with MQTT_DISABLED Device Client was not validating the root-ca file permissions or parent directory permissions. So, this change was intended to not test the file permissions would fail validation if MQTT was disabled. However, realized now that we made root-ca optional and do not set a value unless it is valid then we can validate the file permissions if a value is present regardless of MQTT being disabled.

test/config/TestConfig.cpp
shangabl: I think I used tab in the string literal and it caused a diff between the display in my IDE and the string literal.

marcoemorais-aws (approved):

HarshGandhi-AWS (approved):

BolongZhang-AWS (approved):

PR #266, opened 2022-05-16
Refactor Jobs and add unit tests

shangabl:
Motivation
JobsFeature did not have unit tests. Some refactoring was required for testability.

Modifications
Change summary
Created an interface and wrapper for IotJobsClient so it could be mocked for testing JobsFeature.
Refactored jobs by splitting off methods which create new threads for mocking. Added a method to allow injecting a mock IotJobsClient.
Added unit tests.
Revision diff summary

Testing
Unit tests pass and ran Jobs feature locally.

### DIFF
--- a/source/jobs/IotJobsClientWrapper.cpp
+++ b/source/jobs/IotJobsClientWrapper.cpp
@@ -1,0 +1,78 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "IotJobsClientWrapper.h"
+#include "JobsFeature.h"
+#include "aws/iotjobs/NextJobExecutionChangedSubscriptionRequest.h"
+#include "aws/iotjobs/StartNextPendingJobExecutionRequest.h"
+#include "aws/iotjobs/StartNextPendingJobExecutionSubscriptionRequest.h"
+#include "aws/iotjobs/UpdateJobExecutionRequest.h"
+#include "aws/iotjobs/UpdateJobExecutionSubscriptionRequest.h"
+
+using namespace std;
+using namespace Aws::Iot;
+using namespace Aws::Iot::DeviceClient;
+using namespace Aws::Iot::DeviceClient::Jobs;
+using namespace Aws::Iotjobs;
+
+IotJobsClientWrapper::IotJobsClientWrapper(
+    const std::string &thingName,
+    const std::shared_ptr<Aws::Crt::Mqtt::MqttConnection> connection)
+    : thingName(thingName), mqttConnection(connection)
+{
+    jobsClient = std::unique_ptr<Aws::Iotjobs::IotJobsClient>(new IotJobsClient(mqttConnection));
+}
+void IotJobsClientWrapper::PublishStartNextPendingJobExecution(
+    const StartNextPendingJobExecutionRequest &request,
+    Aws::Crt::Mqtt::QOS qos,
+    const OnPublishComplete &onPubAck)
+{
+    jobsClient->PublishStartNextPendingJobExecution(request, qos, onPubAck);
+}
+void IotJobsClientWrapper::SubscribeToStartNextPendingJobExecutionAccepted(
+    const StartNextPendingJobExecutionSubscriptionRequest &request,
+    Aws::Crt::Mqtt::QOS qos,
+    const OnSubscribeToStartNextPendingJobExecutionAcceptedResponse &handler,
+    const OnSubscribeComplete &onSubAck)
+{
+    jobsClient->SubscribeToStartNextPendingJobExecutionAccepted(request, qos, handler, onSubAck);
+}
+void IotJobsClientWrapper::SubscribeToStartNextPendingJobExecutionRejected(
+    const StartNextPendingJobExecutionSubscriptionRequest &request,
+    Aws::Crt::Mqtt::QOS qos,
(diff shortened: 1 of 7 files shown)


source/jobs/IotJobsClientWrapper.h
```diff
@@ -0,0 +1,119 @@
+#ifndef AWS_IOT_DEVICE_CLIENT_IOTJOBSCLIENTWRAPPER_H
```
RogerZhongAWS, line 1: Don't forget to add the copyright and license statements

source/jobs/JobsFeature.cpp
```diff
@@ -564,7 +567,7 @@ bool JobsFeature::isDuplicateNotification(JobExecutionData job)
     return true;
 }
 
-void JobsFeature::executeJob(JobExecutionData job)
+void JobsFeature::initJob(const JobExecutionData &job)
 {
     LOGM_INFO(TAG, "Executing job: %s", job.JobId->c_str());
```
RogerZhongAWS, line 572: Is this log still needed?

shangabl: Oh yeah good catch, will remove.

source/jobs/IotJobsClientWrapper.h
shangabl: Will do, thanks!

source/jobs/JobsFeature.cpp
```diff
@@ -689,3 +703,8 @@ int JobsFeature::stop()
 
     return 0;
 }
+
+std::shared_ptr<AbstractIotJobsClient> JobsFeature::getJobsClient()
+{
+    return std::shared_ptr<AbstractIotJobsClient>(new IotJobsClientWrapper(thingName, mqttConnection));
+}
```
RogerZhongAWS, line 710: One more thing.. if you don't intend to use call jobsClient from multiple classes or pass that as a parameter, wouldn't it make more sense to use std::unique_ptr?

which then makes this function become 
```suggestion
void JobsFeature::getJobsClient()
{
    jobsClient = std::unique_ptr<AbstractIotJobsClient>(new IotJobsClientWrapper(thingName, mqttConnection));
}
```

and also might want to rename getJobsClient to something like createJobsClient regardless of using unique_ptr or not since it might get confused as a simple getter method while in reality a new object is being created. Feel free to disagree though.

shangabl: The shared_ptr is only because of testing. Since you can't have two unique_ptrs pointing to the same object you can't create a Mock unique_ptr<MockIotJobsClient> and inject it into JobsFeature and then use the mock in the tests. Since unique_ptr is move only you can't keep the reference to it in the tests after injecting it... If that makes sense.

source/jobs/JobsFeature.cpp
```diff
@@ -590,11 +593,22 @@ void JobsFeature::executeJob(JobExecutionData job)
             shutdownHandler);
         return;
     }
-    else
-    {
-        publishUpdateJobExecutionStatus(job, {Iotjobs::JobStatus::IN_PROGRESS, "", "", ""});
-    }
+    publishUpdateJobExecutionStatus(job, {Iotjobs::JobStatus::IN_PROGRESS, "", "", ""});
```
marcoemorais-aws, line 596: IMO: Nicer to create a single argument constructor for `JobExecutionStatusInfo`.

marcoemorais-aws (changes requested):
Good tests, some comments on idiomatic c++ and best practices from core guidelines.

source/jobs/JobsFeature.cpp
shangabl: So, there are several different instantiations of JobExecutionStatusInfo. I can create a single argument constructor for this case but was wondering if there is any reason not to make it single argument with the remaining arguments default "" so that it can be constructed in all the different ways with same constructor (status), (status, "reason"), (status, "", "stdout"), etc.

source/jobs/IotJobsClientWrapper.cpp
```diff
@@ -0,0 +1,78 @@
...
+using namespace Aws::Iot::DeviceClient;
+using namespace Aws::Iot::DeviceClient::Jobs;
+using namespace Aws::Iotjobs;
+
+IotJobsClientWrapper::IotJobsClientWrapper(
+    const std::string &thingName,
+    const std::shared_ptr<Aws::Crt::Mqtt::MqttConnection> connection)
+    : thingName(thingName), mqttConnection(connection)
+{
+    jobsClient = std::unique_ptr<Aws::Iotjobs::IotJobsClient>(new IotJobsClient(mqttConnection));
```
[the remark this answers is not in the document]
shangabl: I did try this initially as you have mentioned told me before to use the member initialization list but it didn't work here and I'm not sure why... Then I forgot to follow up. I get an undefined reference error.

source/jobs/JobsFeature.cpp
marcoemorais-aws: By default std::string will construct empty string. You can create 1-argument (status-only), 2-argument (status,  string), 3-argument (status, string, string) constructors and leave the other strings unset since they will be default constructed to empty string.

shangabl: So, when I went to do this I realized there is never both a stderror and stdout so I made a single argument constructor and then an all-args constructor. Refactored a bit so that when publishing statusDetails it will place the output correctly.

marcoemorais-aws (approved):

RogerZhongAWS (approved):

PR #271, opened 2022-05-18
Disable device defender by default

shangabl:
### Motivation
Disabling Device Defender by default so that customers are not surprised by this feature being enabled without explicitly enabling it.

#### Change summary
Make DD default disabled. Update Config tests to assert DD is disabled with minimum config. Add to documentation.

### Testing
Unit tests

### DIFF
--- a/test/config/TestConfig.cpp
+++ b/test/config/TestConfig.cpp
@@ -71,7 +71,6 @@ class ConfigTestFixture : public ::testing::Test
     {
         ASSERT_TRUE(config.jobs.enabled);
         ASSERT_TRUE(config.tunneling.enabled);
-        ASSERT_TRUE(config.deviceDefender.enabled);
     }
 };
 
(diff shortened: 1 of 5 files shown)


RogerZhongAWS (dismissed):

test/config/TestConfig.cpp
```diff
@@ -217,6 +216,10 @@ TEST_F(ConfigTestFixture, HappyCaseMinimumConfig)
     ASSERT_STREQ(filePath.c_str(), config.key->c_str());
     ASSERT_STREQ("thing-name value", config.thingName->c_str());
     ASSERT_FALSE(config.fleetProvisioning.enabled);
+    ASSERT_FALSE(config.deviceDefender.enabled);
```
marcoemorais-aws, line 219: Don't we have a `ConfigTestFixture::DefaultFeaturesEnabled` that all of these assertions could be rolled up under? Also could you rename that from `ConfigTestFixture::DefaultFeaturesEnabled` to ``ConfigTestFixture::AssertDefaultFeaturesEnabled`?

marcoemorais-aws (approved):
@shangabl One comment take-or-leave it. If you disagree, then ok to merge.

marcoemorais-aws (approved):

RogerZhongAWS (approved):

PR #273, opened 2022-05-20
Add more valgrind suppressions + fixes

RogerZhongAWS:
### Motivation
- Cleaning up valgrind output

#### Change summary
Fixed problems that are for certain related to device client
- invalid memory access in StdOutLogger (use after free)
- minor leak in pubsub
 
Added valgrind suppressions for everything else (direct and indirect leaks from elsewhere)

### Testing
Tested by running valgrind on device client with no features, + jobs, + ST, + DD, + pubsub, + fp, + shadows, + sensor publish
Still not 100% clean, may need followup PR's to address everything.


RogerZhongAWS:
maybe using multiple suppression files would be better..

source/main.cpp
```diff
@@ -352,6 +351,7 @@ int main(int argc, char *argv[])
     memset(&sigset, 0, sizeof(sigset_t));
     int received_signal;
     sigaddset(&sigset, SIGINT);
+    sigaddset(&sigset, SIGTERM);
```
HarshGandhi-AWS, line 354: Duplicate line of code. We already have `sigaddset(&sigset, SIGTERM);` on line 356

RogerZhongAWS: fixed

PR #277, opened 2022-05-31
Create sonarcloud.yml

RogerZhongAWS:
### Motivation
- Add sonarcloud static analysis to our branches.

#### Change summary
- Made a simple change in pubsub to demonstrate.


source/samples/pubsub/PubSubFeature.h
```diff
@@ -89,7 +89,7 @@ namespace Aws
                     /**
                      * \brief Default payload if no publish file was provided
                      */
-                    const std::string DEFAULT_PUBLISH_PAYLOAD = "Hello World!";
+                    const std::string DEFAULT_PUBLISH_PAYLOAD = "{\"Hello\": \"World!\"}";
```
marcoemorais-aws, line 92: Was the intention to convert this to JSON? If so, then better to use a raw string here since that won't require any escaping.  You want to reach for a raw string whenever you need strings that would otherwise require a lot of escaping such as a JSON literal or regex.  See entry labeled `(6)` https://en.cppreference.com/w/cpp/language/string_literal

This is also one of those variables that better as `static const std::string` rather than `const std::string`. See comment here for explanation of the difference https://github.com/awslabs/aws-iot-device-client/pull/278#discussion_r887323906

marcoemorais-aws (changes requested):
A little surprised that sonarcloud didn't find any existing problems. DId you try (for purpose of exercising the service) to introduce a known failure and see whether that results in a build failure (negative test)?

PR #278, opened 2022-06-01
Adds tests for JobEngine and fix stdout/stderr bug in JobsFeature

shangabl:
### Motivation
Need unit tests for JobEngine.
While writing tests I realized a previous change I had made for modifying JobExecutionStatusInfo needed to be fixed. We previously never included both stderr and stdout in the status details but we should have because steps can fail without the job failing.

Still needs more tests but I believe this is a big improvement over having no tests.

### DIFF
--- a/source/jobs/JobsFeature.cpp
+++ b/source/jobs/JobsFeature.cpp
@@ -376,11 +376,11 @@ void JobsFeature::publishUpdateJobExecutionStatus(
         statusDetails["reason"] = statusInfo.reason.substr(0, MAX_STATUS_DETAIL_LENGTH).c_str();
     }
 
-    if (statusInfo.status == Iotjobs::JobStatus::SUCCEEDED && !statusInfo.output.empty())
+    if (!statusInfo.stdoutput.empty())
     {
         // We want the most recent output since we can only include 1024 characters in the job execution update
-        int startPos = statusInfo.output.size() > MAX_STATUS_DETAIL_LENGTH
-                           ? statusInfo.output.size() - MAX_STATUS_DETAIL_LENGTH
+        int startPos = statusInfo.stdoutput.size() > MAX_STATUS_DETAIL_LENGTH
+                           ? statusInfo.stdoutput.size() - MAX_STATUS_DETAIL_LENGTH
                            : 0;
         // TODO We need to add filtering of invalid characters for the status details that may come from weird
         // process output. The valid values for a statusDetail value are '[^\p{C}]+ which translates into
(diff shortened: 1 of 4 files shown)


test/jobs/TestJobEngine.cpp
```diff
@@ -0,0 +1,161 @@
...
+TEST_F(TestJobEngine, ExecuteNoSteps)
+{
+    vector<PlainJobDocument::JobAction> steps;
+    vector<std::string> args;
+    PlainJobDocument jobDocument = createTestJobDocument(steps, true);
+    JobEngine jobEngine;
+
+    int executionStatus = jobEngine.exec_steps(jobDocument, testHandlerDirectoryPath);
+    ASSERT_EQ(executionStatus, 0);
+    ASSERT_TRUE(jobEngine.getStdOut().length() == 0);
```
shangabl, line 159: I'll change this to ASSERT_EQ(..., 0) -- I was going to have it be .empty() then didn't fix here.

source/jobs/JobsFeature.h
```diff
@@ -74,12 +76,13 @@ namespace Aws
                      * \brief The default directory that the Jobs feature will use to find executables matching
                      * an incoming job document's operation attribute
                      */
+                    const std::string DEFAULT_JOBS_HANDLER_DIR = "~/.aws-iot-device-client/jobs/";
```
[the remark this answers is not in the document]
shangabl: Ok cool makes sense. I had just moved this to be under the correct comment. Updated.

shangabl:
I should actually add unit tests to JobsFeature here for the STDOUT/STDERR. Will push shortly.

RogerZhongAWS (approved):

xlcheng1 (approved):

marcoemorais-aws (approved):
These are some of the best tests for any feature we have in the repo 👍

source/jobs/JobsFeature.cpp
```diff
@@ -33,6 +33,8 @@ using namespace Aws::Iot::DeviceClient::Util;
 using namespace Aws::Iot::DeviceClient::Jobs;
 using namespace Aws::Iotjobs;
 
+const std::string JobsFeature::DEFAULT_JOBS_HANDLER_DIR = "~/.aws-iot-device-client/jobs/";
```
BolongZhang-AWS, line 36: Assigned the string value in the header file?

shangabl: This was the comment Marco left about it

```
This should be static const std::string instead of const std::string. The difference is that the former creates exactly 1 instance of the variable no matter how many instances of the class are created and the latter creates an instance of the variable for every instance of the class.

Since std::string is a non integral type you will have to split the declaration and definition.

In .h file add static const std::string DEFAULT_JOBS_HANDLER_DIR;
In .cpp file add const std::string JobsFeature::DEFAULT_JOBS_HANDLER_DIR = "...";
```

marcoemorais-aws: You can only initialize a static const value in the header for integer types, not other types. The second best choice here is to use `static constexpr char[]` or (if c++17) use `static const std::string_view`.

PR #279, opened 2022-06-02
Removed ST feature stop code from connection shutdown callback.

HarshGandhi-AWS:
### Motivation
- We were stopping Device Client ST feature if connection shutdown callback is triggered. We would not like to do that since that will stop feature and then eventually stop DC. It is a functionality bug.

#### Change summary
* Removed feature stop logic. 
* We do not have to worry about calling destructor since `erase` will eventually call object destructor.

#### Revision diff summary

### Testing
- CI test run result: <link>

### DIFF
--- a/source/jobs/JobEngine.h
+++ b/source/jobs/JobEngine.h
@@ -105,6 +105,7 @@ namespace Aws
                         int &executionStatus);
 
                   public:
+                    virtual ~JobEngine() = default;
                     /**
                      * \brief Used by output processing threads to assess output from the child process
                      *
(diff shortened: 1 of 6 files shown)


RogerZhongAWS (approved):

shangabl (approved):

marcoemorais-aws (approved):
Tested this locally by (1) create a tunnel (2) close tunnel (3) create tunnel. Before this commit, the device client would have stopped the secure tunnel feature after step (2) preventing the device client from connecting to the tunnel created in step (3). In addition, if secure tunneling was the only feature enabled, then the device client would shut itself down.

Although I think the new behavior is an improvement since users can create multiple tunnels, I am also wondering if the secure tunneling binary used in GG depends on the original behavior.  We need to check with @shangabl to find out whether the GG secure tunneling component assumes device client shuts itself down after tunnel is closed or whether the GG component sends a signal to device client to initiate shutdown.

HarshGandhi-AWS:
> 

Thank you Marco for testing it.

BolongZhang-AWS (approved):

PR #281, opened 2022-06-07
Resolve Secure Tunneling race condition bug.

HarshGandhi-AWS:
### Motivation
In Device Client, for Secure Tunneling feature, we are seeing a race condition where on 1 thread we are calling destructor to destroy the secure tunnel object and on other thread the call back method for secure tunnel connection close is called. Since the object is destroyed, the call back method is failing to find the object and eventually failing with segmentation fault.

#### Change summary

To resolve the issue, I have updated the code to not to call destructor immediately from the main thread I.E. we are waiting for the callback on connection shutdown to eventually call destructor to destroy the SecureTunnel object. Also it is important to note that the callback and object destruction is done on the same thread.  

#### Revision diff summary

### Testing
- CI test run result: <link>


source/tunneling/SecureTunnelingContext.cpp
```diff
@@ -33,7 +33,16 @@ namespace Aws
                 {
                 }
 
-                SecureTunnelingContext::~SecureTunnelingContext() { mSecureTunnel->Close(); }
+                SecureTunnelingContext::~SecureTunnelingContext() {
+                    LOG_DEBUG(TAG, "SecureTunnelingContext::~SecureTunnelingContext");
+//                    mSecureTunnel->Close();
```
HarshGandhi-AWS, line 38: Hey @marcoemorais-aws , If in call `Close()` over here, I am seeing a segmentation fault. Any thoughts on that?

marcoemorais-aws: I am able to reproduce this locally and looks like we are passing a null pointer to aws_secure_tunnel_close — the Aws::Iotsecuretunneling::SecureTunnel::Close should have a null check.  Please add the snippet below and the issue should be resolved (alternately you can use `mSecureTunnel->IsValid()` as a substitute for `mSecureTunnel->GetUnderlyingHandle()`).

```
                    if (mSecureTunnel->GetUnderlyingHandle())
                    {
                        mSecureTunnel->Close();
                    }
```

marcoemorais-aws (changes requested):
The new shutdown logic works for me and I was able to reproduce and resolve the core dump issue that you mentioned by adding the snippet mentioned in the other comment.

marcoemorais-aws (dismissed):

RogerZhongAWS (approved):

RogerZhongAWS (approved):

BolongZhang-AWS (approved):

PR #287, opened 2022-06-17
Placeholder input validation

xlcheng1:
### Motivation
- If customers accidentally use a placeholder value (or value that cannot be expanded by wordexp) for certain parameters in config file, device client will have a segmentation fault because of ExtractExpandedPath() function.

#### Change summary
- Added checks when parsing config file
- Added a placeholderInput test 

#### Revision diff summary

### Testing
- CI test run result: <link>

### DIFF
--- a/source/config/Config.cpp
+++ b/source/config/Config.cpp
@@ -576,7 +576,7 @@ bool PlainConfig::LogConfig::LoadFromJson(const Crt::JsonView &json)
     jsonKey = JSON_KEY_LOG_FILE;
     if (json.ValueExists(jsonKey))
     {
-        if (!json.GetString(jsonKey).empty())
+        if (!json.GetString(jsonKey).empty() && FileUtils::IsValidFilePath(json.GetString(jsonKey).c_str()))
         {
             deviceClientLogFile = FileUtils::ExtractExpandedPath(json.GetString(jsonKey).c_str());
         }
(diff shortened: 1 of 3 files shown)


source/config/Config.cpp
```diff
@@ -1239,24 +1216,6 @@ bool PlainConfig::PubSub::Validate() const
     {
         return true;
     }
-    if (!publishTopic.has_value() || publishTopic->empty())
```
BolongZhang-AWS, line 1242: Also why is this part removed?

source/config/Config.cpp
```diff
@@ -354,10 +354,6 @@ bool PlainConfig::Validate() const
         LOGM_ERROR(Config::TAG, "*** %s: Certificate is missing ***", DeviceClient::DC_FATAL_ERROR);
         return false;
     }
-    else if (!FileUtils::IsValidFilePath(cert->c_str()))
```
RogerZhongAWS, line 357: I am a little bit confused on this, don't we still need to have this if we pass in params via CLI?

source/config/Config.cpp
```diff
@@ -1047,22 +1040,6 @@ bool PlainConfig::FleetProvisioning::Validate() const
         return false;
     }
 
-    if (csrFile.has_value() && !csrFile->empty())
-    {
-        if (!FileUtils::IsValidFilePath(csrFile->c_str()))
```
RogerZhongAWS, line 1052: here as well

xlcheng1:
closing this pr because PR #288 is providing better approach solving the issue.

Issue #203, opened 2021-12-13
Failure to check return code from `wordexp` in implementation of FileUtils::ExtractExpandedPath


marcoemorais-aws:
**Describe the bug**
In the current implementation of FileUtils::ExtractExpandedPath on main branch (code linked in [1] below), the return code from `wordexp` is not checked. In the event that wordexp returns a non-zero status code, then the `char **we_wordv` member of the `wordexp_t` struct is unset and accessing the first element of `we_wordv` on line 66 will  produce undefined behavior.  

[1] https://github.com/awslabs/aws-iot-device-client/blob/dab360454fb3b8430d63b8fe9547d012d150feae/source/util/FileUtils.cpp#L62-L69

**To Reproduce**

1. Add a handler path with any of the invalid characters `Illegal occurrence of newline or one of |, &, ;, <, >, (, ), {, }.` as described in `WRDE_BADCHAR` of wordexp man page [2].
```
      "jobs":   {
        "enabled":      true,
        "handler-directory": "/tmp/|"
      },
```
2.  Start device client and you will observe the core dump at line 66 of `source/util/FileUtils.cpp`. Stack trace in [3] below.

[2] https://man7.org/linux/man-pages/man3/wordexp.3.html
[3]
```
(gdb) bt
#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:65
#1  0x00007fd5d7f00e9c in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(char const*, std::allocator<char> const&) () from /lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00005603705d3f51 in Aws::Iot::DeviceClient::Util::FileUtils::ExtractExpandedPath (filePath="/tmp/|") at /tmp/aws-iot-device-client-clion/source/util/FileUtils.cpp:66
#3  0x00005603705b3b88 in Aws::Iot::DeviceClient::PlainConfig::Jobs::LoadFromJson (this=0x7ffe14e75680, json=...) at /tmp/aws-iot-device-client-clion/source/config/Config.cpp:618
#4  0x00005603705afee6 in Aws::Iot::DeviceClient::PlainConfig::LoadFromJson (this=0x560370bf7f20 <config>, json=...) at /tmp/aws-iot-device-client-clion/source/config/Config.cpp:131
[...]

PR #288, opened 2022-06-29
Adanfo first CR

adanforth:
### Motivation
- Failure to check return code from wordexp in implementation of FileUtils::ExtractExpandedPath
- Issue number: #203

#### Change summary
Added brief check on the return value of wordexp. Upon receiving a bad status code, return a sentinel value indicating failure. Also added documentation of this behavior on the function description.


source/util/FileUtils.cpp
```diff
@@ -66,7 +66,11 @@ string FileUtils::ExtractExpandedPath(const string &filePath)
         return "";
     }
     wordexp_t word;
-    wordexp(filePath.c_str(), &word, 0); // TODO: How to handle error?
+    int status = wordexp(filePath.c_str(), &word, 0); // TODO: How to handle error?
+    if (status == WRDE_BADCHAR)
+    {
+        return "-1";
```
marcoemorais-aws, line 72: Checking the status and returning a sentinel value prevents the segmentation fault, but since no callers of this method are prepared to check for the returned sentinel this doesn't really address the problem. Furthermore, the sentinel value is potentially ambiguous since "-1" is an unlikely but valid filename path. Instead of returning sentinel value, I recommend that you throw a `std::runtime_error` [1] (or an error with a custom name that inherits from runtime_error as shown in [2]).  The exception message should include the offending `filePath`.

The vast majority of instances in which `FileUtils::ExtractExpandedPath` is called emanate from `Config::init`. As a result, you can put the exception handler in that function. The handler should log the exception message with `LOGM_ERROR` and return `false` from the function. This will fail the device client on startup.

We do have a few instances where `FileUtils::ExtractExpandedPath` is called after config is loaded and while a feature is started. Examples include `SampleShadowFeature.cpp` and `PubSubFeature.cpp`.  It would be better if these feature implementations did the validation as part of `Config::init` so that we can fail the device client while parsing/validating config rather than at feature startup. If that approach isn't possible then you should add try-catch in each feature or possibly as part of `SharedCrtResourceManager::startDeviceClientFeatures` (not sure which is better). 

[1] https://en.cppreference.com/w/cpp/error/runtime_error

[2]
```
struct my_error : std::runtime_error 
{
    using std::runtime_error::runtime_error;
};
```

marcoemorais-aws (changes requested):

source/util/FileUtils.cpp
```diff
@@ -66,7 +66,13 @@ string FileUtils::ExtractExpandedPath(const string &filePath)
         return "";
     }
     wordexp_t word;
-    wordexp(filePath.c_str(), &word, 0); // TODO: How to handle error?
+    int status = wordexp(filePath.c_str(), &word, 0); // TODO: How to handle error?
+    if (status)
+    {
+        throw Aws::Iot::DeviceClient::Util::wordexp_fail_error(
+            "Recieved status from wordexp: " + std::to_string(status) +
```
shangabl, line 73: Received is misspelled and I don't think you'd want a new line here.

shangabl (commented):
Can you add some unit tests for Config as well?

source/util/FileUtils.cpp
adanforth: Fixed typo! The new line is to keep the line was added by linter

source/config/Config.cpp
```diff
@@ -2116,55 +2213,64 @@ bool Config::ParseCliArgs(int argc, char **argv, CliArgs &cliArgs)
...
-    if (ParseConfigFile(Config::DEFAULT_FLEET_PROVISIONING_RUNTIME_CONFIG_FILE, true) &&
-        ValidateAndStoreRuntimeConfig())
+    catch (const std::exception &e)
     {
-        LOGM_INFO(
-            TAG,
-            "Successfully fetched Runtime config file '%s' and validated its content.",
-            Config::DEFAULT_FLEET_PROVISIONING_RUNTIME_CONFIG_FILE);
-    }
+        LOGM_ERROR(TAG, "Error while parsing filepath", e.what());
```
marcoemorais-aws, line 2270: The purpose taking a generic `std::exception` and `what()` message is to handle any kind of error generically. The explicit message `Error while parsing filepath` might not correspond to all circumstances raising an exception. Please make the message generic such as `"Error while initializing runtime configuration: %s", e.what()`,

source/fleetprovisioning/FleetProvisioning.cpp
```diff
@@ -489,92 +489,103 @@ bool FleetProvisioning::RegisterThing(Iotidentity::IotIdentityClient identityCli
...
-        if (!ExportRuntimeConfig(
-                Config::DEFAULT_FLEET_PROVISIONING_RUNTIME_CONFIG_FILE,
-                certPath.c_str(),
-                keyPath.c_str(),
-                thingName.c_str(),
-                deviceConfig.c_str()))
-        {
-            return false;
-        }
+        LOGM_ERROR(TAG, "Error while parsing filepath", e.what());
```
marcoemorais-aws, line 585: See earlier comment about making the message generic.

source/main.cpp
```diff
@@ -197,33 +197,42 @@ void handle_feature_stopped(const Feature *feature)
...
+                return false;
+            }
+        };
+        std::thread attemptConnectionThread(
+            [retryConfig, publishLambda] { Retry::exponentialBackoff(retryConfig, publishLambda); });
+        attemptConnectionThread.join();
+    }
+    catch (const std::exception &e)
+    {
+        LOGM_ERROR(TAG, "Error while parsing path", e.what());
```
marcoemorais-aws, line 232: See earlier comment about making the message generic.

source/main.cpp
```diff
@@ -197,33 +197,42 @@ void handle_feature_stopped(const Feature *feature)
...
+        LoggerFactory::getLoggerInstance()->shutdown();
+        deviceClientAbort("Invalid configuration");
```
marcoemorais-aws, line 234: Change message to `Failure from attemptConnection`.

source/util/FileUtils.cpp
```diff
@@ -66,7 +66,13 @@ string FileUtils::ExtractExpandedPath(const string &filePath)
         return "";
     }
     wordexp_t word;
-    wordexp(filePath.c_str(), &word, 0); // TODO: How to handle error?
+    int status = wordexp(filePath.c_str(), &word, 0); // TODO: How to handle error?
+    if (status)
+    {
+        throw Aws::Iot::DeviceClient::Util::wordexp_fail_error(
+            "Received status from wordexp: " + std::to_string(status) +
+            "\nPertaining to following filepath: " + filePath);
```
marcoemorais-aws, line 74: No embedded newlines in the message since this will mess up the logging which prepends a severity (`DEBUG|INFO|WARN|ERROR`) to every line in the log file.

source/util/FileUtils.cpp
```diff
@@ -66,7 +66,13 @@ string FileUtils::ExtractExpandedPath(const string &filePath)
         return "";
     }
     wordexp_t word;
-    wordexp(filePath.c_str(), &word, 0); // TODO: How to handle error?
+    int status = wordexp(filePath.c_str(), &word, 0); // TODO: How to handle error?
+    if (status)
+    {
+        throw Aws::Iot::DeviceClient::Util::wordexp_fail_error(
```
marcoemorais-aws, line 72: Minor: You prolly don't need the full namespace now. If you still see the error create a shorter synonym with a using declaration. https://en.cppreference.com/w/cpp/language/using_declaration

test/util/TestFileUtils.cpp
```diff
@@ -68,6 +68,19 @@ TEST(FileUtils, handlesEmptyPathToExtractExpandedPath)
     ASSERT_STREQ("", extendedPath.c_str());
 }
 
+TEST(FileUtils, handlesIllegalCharacterInExtractExpandedPath)
+{
+    ASSERT_THROW(FileUtils::ExtractExpandedPath("|"), wordexp_fail_error);
```
marcoemorais-aws, line 73: Google test provides support for Value-Parameterized tests which are a better match for this type of test. https://github.com/google/googletest/blob/main/docs/advanced.md#value-parameterized-tests

marcoemorais-aws (changes requested):
Direction is good. Some bugs and some style recommendations.

source/util/FileUtils.cpp
shangabl: I meant the "\n" in the output. No reason for that I think.

adanforth: resolved

source/config/Config.cpp
```diff
@@ -1404,6 +1480,26 @@ bool PlainConfig::SampleShadow::Validate() const
...
+        if (!FileUtils::FileExists(shadowInputFile->c_str()))
+        {
+            LOGM_ERROR(
+                Config::TAG,
+                "Unable to open shadow data file %s, file does not exist",
+                Sanitize(shadowInputFile->c_str()).c_str());
+            return false;
+        }
+        size_t incomingFileSize = FileUtils::GetFileSize(shadowInputFile->c_str());
+        if (4 * 1024 < incomingFileSize)
```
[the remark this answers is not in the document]
adanforth: Actually unsure of the limit. Who would know about its origin?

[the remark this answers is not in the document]
marcoemorais-aws: I believe that intended limit is the maximum size of shadow state document which is currently 8kb (search for `Maximum size of a JSON state document`). Please check with @HarshGandhi-AWS and if the intent was to enforce the shadow limit, then let's make the constant `8*1024`.

https://docs.aws.amazon.com/general/latest/gr/iot-core.html#device-shadow-limits

test/config/TestConfig.cpp
```diff
@@ -211,7 +211,12 @@ TEST_F(ConfigTestFixture, AllFeaturesEnabled)
     config.sampleShadow.SerializeToObject(sampleShadow);
     ASSERT_STREQ("shadow-name", sampleShadow.View().GetString(config.sampleShadow.JSON_SAMPLE_SHADOW_NAME).c_str());
     ASSERT_STREQ("", sampleShadow.View().GetString(config.sampleShadow.JSON_SAMPLE_SHADOW_INPUT_FILE).c_str());
-    ASSERT_STREQ("", sampleShadow.View().GetString(config.sampleShadow.JSON_SAMPLE_SHADOW_OUTPUT_FILE).c_str());
+    ostringstream outputPathStream;
+    outputPathStream << Config::DEFAULT_SAMPLE_SHADOW_OUTPUT_DIR << Config::DEFAULT_SAMPLE_SHADOW_DOCUMENT_FILE;
```
marcoemorais-aws, line 215: DRY. Make `outputPath` a member variable of the `ConfigTestFixture`, initialize and extract path in `ConfigTestFixture::SetUp()`, and then you can reference it inside in any test that uses the fixture (tests with `TEST_F` prefix).

marcoemorais-aws (commented):
Looks great, just a few minor comments with the test. I will try to use it as well.

source/config/Config.cpp
```diff
@@ -2116,55 +2189,72 @@ bool Config::ParseCliArgs(int argc, char **argv, CliArgs &cliArgs)
...
-    }
+        if (ParseConfigFile(Config::DEFAULT_FLEET_PROVISIONING_RUNTIME_CONFIG_FILE, true) &&
+            ValidateAndStoreRuntimeConfig())
+        {
+            LOGM_INFO(
+                TAG,
+                "Successfully fetched Runtime config file '%s' and validated its content.",
+                Config::DEFAULT_FLEET_PROVISIONING_RUNTIME_CONFIG_FILE);
+        }
+        else
```
marcoemorais-aws, line 2241: I misguided you on this comment, please remove the `else` branch here.

source/config/Config.cpp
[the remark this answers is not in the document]
HarshGandhi-AWS: @adanforth  I agree with Marco on this, file size limit should be `8 Kilobytes`. Add a note in Shadow feature documentation regarding the limit and also mention that the limit is adjustable.

HarshGandhi-AWS (changes requested):

marcoemorais-aws (approved):

shangabl (dismissed):

HarshGandhi-AWS (dismissed):
Update the PR title and final commit message to something more relevant. Everything else looks good to me.

marcoemorais-aws (dismissed):
Nice work and congrats on your first commit!

marcoemorais-aws (approved):

shangabl (approved):

RogerZhongAWS (approved):

PR #300, opened 2022-07-22
Disable sensor publish feature for secure tunnel builds

marcoemorais-aws:
### Motivation
- Exclude sensor publish feature when building device client binary exclusively for secure tunnel eg `st_component_mode`.

### Modifications
- When `EXCLUDE_SENSOR_PUBLISH=ON` then disable sensor publish configuration validation checks in unit tests.

### Testing
- Unit tests.

### DIFF
--- a/test/config/TestConfig.cpp
+++ b/test/config/TestConfig.cpp
@@ -1258,7 +1258,9 @@ TEST_F(ConfigTestFixture, SensorPublishInvalidConfigAddr)
     PlainConfig config;
     config.LoadFromJson(jsonView);
 
+#if !defined(EXCLUDE_SENSOR_PUBLISH)
     ASSERT_FALSE(config.Validate()); // Invalid permissions on addr.
+#endif
     ASSERT_TRUE(config.sensorPublish.enabled);
     ASSERT_EQ(config.sensorPublish.settings.size(), 1);
     const auto &settings = config.sensorPublish.settings[0];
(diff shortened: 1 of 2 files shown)


test/config/TestConfig.cpp
```diff
@@ -1456,7 +1467,9 @@ TEST_F(ConfigTestFixture, SensorPublishDisableFeature)
     PlainConfig config;
     config.LoadFromJson(jsonView);
 
+#if !defined(EXCLUDE_SENSOR_PUBLISH)
```
HarshGandhi-AWS, line 1470: I was thinking of excluding Sensor publish tests completely instead of only excluding the assert statements.

rushilxc (dismissed):

test/config/TestConfig.cpp
```diff
@@ -1258,6 +1258,9 @@ TEST_F(ConfigTestFixture, SensorPublishInvalidConfigAddr)
...
```
HarshGandhi-AWS, line 1261: shouldn't this be like this without negation `!`?

```
#if defined(EXCLUDE_SENSOR_PUBLISH)
    GTEST_SKIP();
#endif
```

marcoemorais-aws: Dooh! Thank you @HarshGandhi-AWS

rushilxc (approved):

PR #302, opened 2022-07-25
Ignore correct docs folder, and update readme.

JustinBerger:
### Motivation
Noticed that gitignore and readme both referenced doc folder, when it's actually docs.

#### Change summary
Fix gitignore and readme to use docs instead of doc
#### Revision diff summary

### Testing
Minimally tested. No change to code files.
Follow readme
cd docs
docygen
generated files don't show up in git changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,8 @@ out/
 build/
 
 # Ignore compiled documentation folders
-doc/html/
-doc/latex/
+docs/html/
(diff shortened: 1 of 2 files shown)


RogerZhongAWS (approved):

HarshGandhi-AWS (approved):
Thank you for making these changes. Changes looks good to me.

PR #304, opened 2022-07-26
Adanfo feature registry

adanforth:
### Motivation
Currently, main.cpp is using a vector to keep track of its active features. This is a very cumbersome solution. We introduce a new class, FeatureRegistry, which will behave exactly the same, but will abstract away some of the logic into its methods.

#### Change summary
- Added FeatureRegistry.h/.cpp
- Added TestFeatureRegistry.cpp
- Modified Main.cpp to make use of this new class
- Modified SharedCrtResourceManager to make use of this class as well

### Testing
Small Unit tests have been created to test basic functionality

### NOTE
ConfigShadow Feature currently has an error when enabled, the DC will not shut down gracefully.

### DIFF
--- a/source/FeatureRegistry.cpp
+++ b/source/FeatureRegistry.cpp
@@ -1,0 +1,85 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "FeatureRegistry.h"
+#include "logging/LoggerFactory.h"
+
+using namespace std;
+using namespace Aws::Iot::DeviceClient;
+using namespace Aws::Iot::DeviceClient::Util;
+using namespace Aws::Iot::DeviceClient::Logging;
+
+constexpr char FeatureRegistry::TAG[];
+
+std::shared_ptr<Feature> FeatureRegistry::get(const std::string &name) const
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+    if (features.count(name))
+    {
+        return features.at(name);
+    }
+    LOGM_WARN(TAG, "Feature, %s, not found in current registry", name.c_str());
+    return nullptr;
+}
+
+void FeatureRegistry::add(std::string name, std::shared_ptr<Feature> featureToAdd)
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+    if (!features.count(name))
+    {
+        features[name] = featureToAdd;
+    }
+    else
+    {
+        LOGM_WARN(TAG, "Attemped to add %s to Feature Registry despite it already existing", name.c_str());
+    }
+}
+
+// NOTE{adanforth} disable is being called in TestFeatureRegistry, and will also be tested later down the line with
+// the dynamic reload project
+// cppcheck-suppress unusedFunction
+void FeatureRegistry::disable(const std::string &name)
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+    if (features.count(name))
+    {
+        features[name] = nullptr;
+    }
+}
+
+std::size_t FeatureRegistry::getSize() const
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+    return count_if(features.begin(), features.end(), [](const std::pair<string, std::shared_ptr<Feature>> feature) {
+        return feature.second != nullptr;
+    });
+}
+
+void FeatureRegistry::stopAll()
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+
+    for (auto feature : features)
+    {
+        if (feature.second != nullptr)
+        {
+            LOGM_DEBUG(TAG, "Attempting to stop %s", feature.first.c_str());
+            feature.second->stop();
+            features[feature.first] = nullptr;
+        }
+    }
+}
+
+void FeatureRegistry::startAll() const
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+
+    for (auto feature : features)
+    {
+        if (feature.second != nullptr)
+        {
+            LOGM_DEBUG(TAG, "Attempting to start %s", feature.first.c_str());
+            feature.second->start();
+        }
+    }
+}
--- a/source/FeatureRegistry.cpp
+++ b/source/FeatureRegistry.cpp
@@ -22,7 +22,7 @@ std::shared_ptr<Feature> FeatureRegistry::get(const std::string &name) const
     return nullptr;
 }
 
-void FeatureRegistry::add(std::string name, std::shared_ptr<Feature> featureToAdd)
+void FeatureRegistry::add(const std::string &name, std::shared_ptr<Feature> featureToAdd)
 {
     std::lock_guard<std::mutex> lock(featuresLock);
     if (!features.count(name))
--- a/source/SharedCrtResourceManager.h
+++ b/source/SharedCrtResourceManager.h
@@ -5,12 +5,15 @@
 #define DEVICE_CLIENT_SHAREDCRTRESOURCEMANAGER_H
 
 #include "Feature.h"
+#include "FeatureRegistry.h"
 #include "config/Config.h"
 
 #include <atomic>
 #include <aws/crt/Api.h>
 #include <aws/iot/MqttClient.h>
 #include <iostream>
+
+using namespace Aws::Iot::DeviceClient::Util;
 
 namespace Aws
 {
@@ -42,7 +45,7 @@ namespace Aws
                 std::shared_ptr<Crt::Mqtt::MqttConnection> connection;
                 aws_allocator *allocator{nullptr};
                 aws_mem_trace_level memTraceLevel{AWS_MEMTRACE_NONE};
-                std::vector<Feature *> *features;
+                std::shared_ptr<FeatureRegistry> features;
 
                 bool setupLogging(const PlainConfig &config);
 
@@ -73,7 +76,7 @@ namespace Aws
                 static const int RETRY = 1;
                 static const int ABORT = 2;
 
-                bool initialize(const PlainConfig &config, std::vector<Feature *> *features);
+                bool initialize(const PlainConfig &config, std::shared_ptr<Util::FeatureRegistry> featureRegistry);
 
                 void initializeAWSHttpLib();
 
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -147,6 +147,7 @@ void shutdown()
     resourceManager->disconnect();
 #endif
     LoggerFactory::getLoggerInstance().get()->shutdown();
+    resourceManager.reset();
     exit(0);
 }
 
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -3,6 +3,7 @@
 
 #include "ClientBaseNotification.h"
 #include "Feature.h"
+#include "FeatureRegistry.h"
 #include "SharedCrtResourceManager.h"
 #include "Version.h"
 #include "config/Config.h"
@@ -94,10 +95,9 @@ using namespace Aws::Iot::DeviceClient::SensorPublish;
 
 const char *TAG = "Main.cpp";
 
-vector<Feature *> features;
+shared_ptr<FeatureRegistry> features;
 shared_ptr<SharedCrtResourceManager> resourceManager;
 unique_ptr<LockFile> lockFile;
-mutex featuresReadWriteLock;
 bool attemptingShutdown{false};
 Config config;
 
@@ -131,32 +131,23 @@ bool init(int argc, char *argv[])
  */
 void shutdown()
 {
-    featuresReadWriteLock.lock(); // LOCK
-    // Make a copy of the features vector for thread safety
-    vector<Feature *> featuresCopy = features;
-    featuresReadWriteLock.unlock(); // UNLOCK
-
-    if (!attemptingShutdown && !featuresCopy.empty())
+    LOG_DEBUG(TAG, "Inside of shutdown()");
+    if (!attemptingShutdown && features->getSize() != 0)
     {
         attemptingShutdown = true;
 
-        for (auto &feature : featuresCopy)
-        {
-            LOGM_DEBUG(TAG, "Attempting shutdown of %s", feature->getName().c_str());
-            feature->stop();
-        }
-
-        resourceManager->dumpMemTrace();
+        LOG_DEBUG(TAG, "Calling stop all");
+        features->stopAll();
     }
-    else
-    {
+    resourceManager->dumpMemTrace();
+
+    LOG_INFO(TAG, "All features have stopped");
 // terminate program
 #if !defined(DISABLE_MQTT)
-        resourceManager.get()->disconnect();
+    resourceManager->disconnect();
 #endif
-        LoggerFactory::getLoggerInstance().get()->shutdown();
-        exit(0);
-    }
+    LoggerFactory::getLoggerInstance().get()->shutdown();
+    exit(0);
 }
 
 /**
@@ -169,30 +160,6 @@ void deviceClientAbort(const string &reason)
     cout << "AWS IoT Device Client must abort execution, reason: " << reason << endl;
     cout << "Please check the AWS IoT Device Client logs for more information" << endl;
     exit(EXIT_FAILURE);
-}
-
-void handle_feature_stopped(const Feature *feature)
-{
-    featuresReadWriteLock.lock(); // LOCK
-
-    for (int i = 0; (unsigned)i < features.size(); i++)
-    {
-        if (features.at(i) == feature)
-        {
-            // Performing bookkeeping so we know when all features have stopped
-            // and the entire program can be shutdown
-            features.erase(features.begin() + i);
-        }
-    }
-
-    const int size = features.size();
-    featuresReadWriteLock.unlock(); // UNLOCK
-
-    if (0 == size)
-    {
-        LOG_INFO(TAG, "All features have stopped");
-        shutdown();
-    }
 }
 
 void attemptConnection()
@@ -259,7 +226,6 @@ namespace Aws
                         case ClientBaseEventNotification::FEATURE_STOPPED:
                         {
                             LOGM_INFO(TAG, "%s has stopped", feature->getName().c_str());
-                            handle_feature_stopped(feature);
                             break;
                         }
                         default:
@@ -351,6 +317,7 @@ int main(int argc, char *argv[])
     {
         return -1;
     }
+    features = make_shared<FeatureRegistry>();
 
     LOGM_INFO(TAG, "Now running AWS IoT Device Client version %s", DEVICE_CLIENT_VERSION_FULL);
 
@@ -365,7 +332,7 @@ int main(int argc, char *argv[])
     shared_ptr<DefaultClientBaseNotifier> listener =
         shared_ptr<DefaultClientBaseNotifier>(new DefaultClientBaseNotifier);
     resourceManager = shared_ptr<SharedCrtResourceManager>(new SharedCrtResourceManager);
-    if (!resourceManager.get()->initialize(config.config, &features))
+    if (!resourceManager.get()->initialize(config.config, features))
     {
         LOGM_ERROR(TAG, "*** %s: Failed to initialize AWS CRT SDK.", DC_FATAL_ERROR);
         LoggerFactory::getLoggerInstance()->shutdown();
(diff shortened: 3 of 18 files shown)


source/SharedCrtResourceManager.h
```diff
@@ -42,7 +43,8 @@ namespace Aws
                 std::shared_ptr<Crt::Mqtt::MqttConnection> connection;
                 aws_allocator *allocator{nullptr};
                 aws_mem_trace_level memTraceLevel{AWS_MEMTRACE_NONE};
-                std::vector<Feature *> *features;
+                Util::FeatureRegistry *features;
```
marcoemorais-aws, line 46: We want to avoid naked pointers except for C runtime objects such as `aws_allocator` and so forth. Make this a shared pointer to `FeatureRegistry`.

```
std::shared_ptr<Util::FeatureRegistry> features;
```

source/SharedCrtResourceManager.h
```diff
@@ -73,7 +75,7 @@ namespace Aws
                 static const int RETRY = 1;
                 static const int ABORT = 2;
 
-                bool initialize(const PlainConfig &config, std::vector<Feature *> *features);
+                bool initialize(const PlainConfig &config, Util::FeatureRegistry *featureRegistry);
```
marcoemorais-aws, line 78: Take a shared pointer `std::shared_ptr<Util::FeatureRegistry> features`.

source/main.cpp
```diff
@@ -94,10 +95,11 @@ using namespace Aws::Iot::DeviceClient::SensorPublish;
 
 const char *TAG = "Main.cpp";
 
-vector<Feature *> features;
+// vector<Feature *> features;
```
marcoemorais-aws, line 98: Don't leave commented code like this since it will be confusing and more difficult to read. I won't repeat that but it applies in the code that follows. Since we are using version control, we can always examine the previous code without needing to leave it commented in the source file.

source/main.cpp
```diff
@@ -94,10 +95,11 @@ using namespace Aws::Iot::DeviceClient::SensorPublish;
...
+FeatureRegistry features;
```
marcoemorais-aws, line 99: Make it a `shared_ptr<FeatureRegistry>` and similar to `SharedCrtResourceManager` initialize in main.cpp.

source/main.cpp
```diff
@@ -456,7 +463,8 @@ int main(int argc, char *argv[])
         LOG_INFO(TAG, "Jobs is enabled");
         jobs = unique_ptr<JobsFeature>(new JobsFeature());
         jobs->init(resourceManager->getConnection(), listener, config.config);
-        features.push_back(jobs.get());
+        features.add(shared_ptr<Feature>(jobs.get()));
```
marcoemorais-aws, line 466: Bug. unique_ptr is about transfer of ownership. When you call `.get()` on a unique pointer you get a raw pointer to the pointed to object, but the ownership of the object it points to remains with the original unique_ptr. If you were to pass the unique_ptr to FeatureRegistry then the ownership transfers and you can't refer to that object. 

Instead of using `.get()`, what you want to do here is change all of these `unique_ptr` Feature instances to `shared_ptr` and then pass the shared_ptr to `FeatureRegistry::add`. Both main.cpp and FeatureRegistry will share the reference to the same memory. That memory will be deallocated when the reference count for each goes to 0.

marcoemorais-aws (changes requested):

source/main.cpp
```diff
@@ -568,6 +580,9 @@ int main(int argc, char *argv[])
                 break;
             case SIGHUP:
                 resourceManager->dumpMemTrace();
+                FeatureRegistry reg;
```
marcoemorais-aws, line 583: This is weird. You have to call stop on the registry with features in it, not this temporary.

marcoemorais-aws (changes requested):

source/main.cpp
adanforth: That was left overs from testing, I forgot to delete in the push.

source/util/FeatureRegistry.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    LOGM_WARN(TAG, "Feature, %s, not found in current registry", name.c_str());
+
+    return nullptr;
+}
+
+void FeatureRegistry::add(std::shared_ptr<Feature> feature)
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+    if (std::find(features.begin(), features.end(), feature) != features.end())
```
marcoemorais-aws, line 32: This loop will search for features using `operator==`. In this case that means feature is considered equal when address are equal. I am not sure whether that is what you intended and the behavior of `add` is not symmetric with `get` since the latter matches by name. If you instead wanted to search for feature matching by name, then you need something like this.

```
auto result = std::find_if(
    std::begin(features), 
    std::end(features), 
    [&feature](auto f) { return f->getName() == feature->getName(); }
);
```

source/util/FeatureRegistry.cpp
```diff
@@ -0,0 +1,78 @@
...
+    {
+        if (features.at(i)->getName() == name)
+        {
+            features.erase(features.begin() + i);
+            break;
+        }
+    }
+}
+
+int FeatureRegistry::getSize()
```
marcoemorais-aws, line 55: 1. Return type of `std::vector::size` is `std::size_t` (or `std::vector::size_type` which is same thing). Similar to the previous comment, use `std::size_t` instead of `int`.
2. Since this class maintains a lock to protect concurrent modification of `features`, then this function also needs to acquire a lock before invoking member function on `features`.

source/util/FeatureRegistry.cpp
```diff
@@ -0,0 +1,78 @@
...
+{
+    return features.size();
+}
+
+void FeatureRegistry::stopAll()
+{
+    while (!features.empty())
```
marcoemorais-aws, line 62: This logic is strange. Assuming there are no side-effects it will just call `Feature::stop` on the same index in an endless loop.  Based on the name I would expect that this function is symmetric with `startAll` and iterates through all the features and calls `Feature::stop`. If the intent is to both `stop` a feature and do something else (I think that it is), that seems flawed and let's talk about that offline.

Similarly strange is the fact that `startAll` acquires a lock but `stopAll` does not.

source/util/FeatureRegistry.h
```diff
@@ -0,0 +1,49 @@
...
+                 * @brief A class to keep track of the features currently running on the device client
+                 */
+                class FeatureRegistry
+                {
+                  public:
+                    FeatureRegistry() = default;
+                    std::shared_ptr<Feature> get(const std::string &name) const;
+                    void add(std::shared_ptr<Feature> feature);
+                    void remove(const std::string &name);
+                    int getSize();
```
marcoemorais-aws, line 34: `getSize()`, `stopAll`, and `startAll` can all be `const` member functions eg add `const` suffix.

test/util/TestFeatureRegistry.cpp
```diff
@@ -0,0 +1,44 @@
...
+using namespace std;
+using namespace Aws::Iot::DeviceClient;
+using namespace Aws::Iot::DeviceClient::Util;
+using namespace Aws::Iot::DeviceClient::DeviceDefender;
+
+class TestFeatureRegistry : public ::testing::Test
+{
+  public:
+    shared_ptr<FeatureRegistry> features;
+    shared_ptr<DeviceDefenderFeature> deviceDefender;
```
marcoemorais-aws, line 20: You can invent dummy features for this test. Use something like to below to create `FakeFeature` instances with different names so that you can test the api with more than 1 entry in the registry. Overall you need more tests.

```
class FakeFeature : public Feature
{
  public:
    FakeFeature(const std::string& name) :name(name) {}
    void start() override { }
    void stop() override { }
    std::string getName() { return name; }
  private:
     std::string name;
};
```

source/main.cpp
```diff
@@ -447,16 +420,16 @@ int main(int argc, char *argv[])
...
+    // featuresReadWriteLock.lock(); // LOCK
 
 #if !defined(EXCLUDE_JOBS)
-    unique_ptr<JobsFeature> jobs;
+    shared_ptr<JobsFeature> jobs;
     if (config.config.jobs.enabled)
     {
         LOG_INFO(TAG, "Jobs is enabled");
-        jobs = unique_ptr<JobsFeature>(new JobsFeature());
+        jobs = shared_ptr<JobsFeature>(new JobsFeature());
```
marcoemorais-aws, line 430: Be consistent and use `make_shared` throughout this file.

source/SharedCrtResourceManager.h
```diff
@@ -42,7 +43,8 @@ namespace Aws
                 std::shared_ptr<Crt::Mqtt::MqttConnection> connection;
                 aws_allocator *allocator{nullptr};
                 aws_mem_trace_level memTraceLevel{AWS_MEMTRACE_NONE};
-                std::vector<Feature *> *features;
+                std::shared_ptr<Util::FeatureRegistry> features;
+                // std::vector<Feature *> *features;
```
marcoemorais-aws, line 47: Remove commented code, we have version control. I see this in other places in this review, but won't repeat comment.

source/util/FeatureRegistry.h
```diff
@@ -0,0 +1,49 @@
...
+
+                /**
+                 * @brief A class to keep track of the features currently running on the device client
+                 */
+                class FeatureRegistry
+                {
+                  public:
+                    FeatureRegistry() = default;
+                    std::shared_ptr<Feature> get(const std::string &name) const;
+                    void add(std::shared_ptr<Feature> feature);
```
marcoemorais-aws, line 32: Add comments to describe each function.

marcoemorais-aws (changes requested):

source/SharedCrtResourceManager.h
```diff
@@ -42,7 +43,7 @@ namespace Aws
                 std::shared_ptr<Crt::Mqtt::MqttConnection> connection;
                 aws_allocator *allocator{nullptr};
                 aws_mem_trace_level memTraceLevel{AWS_MEMTRACE_NONE};
-                std::vector<Feature *> *features;
+                std::shared_ptr<Util::FeatureRegistry> features;
```
shangabl, line 46: nit: use Util namespace so you can just have <FeatureRegistry>

source/main.cpp
```diff
@@ -94,10 +95,9 @@ using namespace Aws::Iot::DeviceClient::SensorPublish;
 
 const char *TAG = "Main.cpp";
 
-vector<Feature *> features;
+shared_ptr<FeatureRegistry> features = make_shared<FeatureRegistry>();
```
shangabl, line 98: Just declare here and then init in the main method.

source/main.cpp
```diff
@@ -131,32 +131,24 @@ bool init(int argc, char *argv[])
...
 
-        for (auto &feature : featuresCopy)
-        {
-            LOGM_DEBUG(TAG, "Attempting shutdown of %s", feature->getName().c_str());
-            feature->stop();
-        }
-
-        resourceManager->dumpMemTrace();
+        LOG_DEBUG(TAG, "Calling stop all");
+        features->stopAll();
```
shangabl, line 140: Why not just remove each feature after you stop it? Seems like these are only called together.

source/util/FeatureRegistry.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "FeatureRegistry.h"
+#include "../logging/LoggerFactory.h"
+
+using namespace Aws::Iot::DeviceClient;
+using namespace Aws::Iot::DeviceClient::Util;
+using namespace Aws::Iot::DeviceClient::Logging;
+
+constexpr char FeatureRegistry::TAG[];
+
+std::shared_ptr<Feature> FeatureRegistry::get(const std::string &name) const
```
shangabl, line 13: Is this used? I'm not seeing it used outside of your unit tests.

source/main.cpp
adanforth: There were some logic issues doing it that way, which is how it was originally done. the call to remove requires the lock as well, recursive locks could be used in that case, but I was advised against them. I had issues with the existing logic inside of included handle_feature_stopped, which would call shutdown once the vector was empty which means we couldn't remove the last feature since now we track the features in a vector in FeatureRegistry. However, now that I removed that logic, we could just move the logic of this removeAll() call inside of stopAll(), but I don't see a need to currently. 
To summarize, I made it this way to work around other errors in main.cpp, but those errors are gone and now this remains, but I think keeping it this way is fine.

source/util/FeatureRegistry.cpp
adanforth: I will be using get in dynamic reload

source/main.cpp
shangabl: Adding an entire new method to a class because of something that no longer exists is generally not recommended.

source/util/FeatureRegistry.cpp
shangabl: I would probably just add it in your dynamic reload PR then. Adding a method you will need later that doesn't actually get used does not seem like a good idea. What if AppSec doesn't approve your project or it is never merged for some other reason?

marcoemorais-aws: This method is prolly needed for testing.

source/FeatureRegistry.cpp
```diff
@@ -0,0 +1,90 @@
...
+using namespace Aws::Iot::DeviceClient::Util;
+using namespace Aws::Iot::DeviceClient::Logging;
+
+constexpr char FeatureRegistry::TAG[];
+
+std::shared_ptr<Feature> FeatureRegistry::get(const std::string &name) const
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+
+    for (auto &feature : features)
```
shangabl: Addressed in Slack

source/FeatureRegistry.cpp
```diff
@@ -0,0 +1,90 @@
...
+    LOGM_WARN(TAG, "Feature, %s, not found in current registry", name.c_str());
+
+    return nullptr;
+}
+
+void FeatureRegistry::add(std::string name, std::shared_ptr<Feature> featureToAdd)
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+
+    for (auto &feature : features)
```
shangabl: same

source/FeatureRegistry.cpp
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    features[name] = featureToAdd;
+}
+
+void FeatureRegistry::remove(const std::string &name)
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+    for (auto &feature : features)
```
shangabl: same

source/FeatureRegistry.cpp
```diff
@@ -0,0 +1,90 @@
...
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+    return features.size();
+}
+
+void FeatureRegistry::stopAll()
+{
+    std::lock_guard<std::mutex> lock(featuresLock);
+
+    for (auto feature : features)
```
shangabl, line 57: You shouldn't need two for-loops here.

test/util/TestFeatureRegistry.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    void SetUp() override
+    {
+        features = make_shared<FeatureRegistry>();
+        feature1 = make_shared<FakeFeature>("feature-1");
+        feature2 = make_shared<FakeFeature>("feature-2");
+        feature3 = make_shared<FakeFeature>("feature-3");
+    }
+
+    void TearDown() override { features = make_shared<FeatureRegistry>(); }
```
shangabl, line 56: I don't think this is needed? Could be wrong though.

the features shared_ptr will be reinitialized before each test by the SetUp method.

test/util/TestFeatureRegistry.cpp
```diff
@@ -0,0 +1,132 @@
...
+};
+
+TEST_F(TestFeatureRegistry, AddFeaturesToRegistry)
```
shangabl, line 59: Add comments to explain each test.

Also, I think it would be better to have each test only doing one thing. This is testing to make sure add is idempotent and that multiple distinct features can be added and that the size will be correct... I think you'd actually want to test that the size was correct and that it contained all the correct features. But perhaps split the idempotency test off?

test/util/TestFeatureRegistry.cpp
```diff
@@ -0,0 +1,132 @@
...
+    ASSERT_EQ(1, features->getSize());
+
+    features->add(feature2->getName(), feature2);
+    ASSERT_EQ(2, features->getSize());
+
+    features->add(feature3->getName(), feature3);
+    ASSERT_EQ(3, features->getSize());
+}
+
+TEST_F(TestFeatureRegistry, GetNonExistentFeature)
```
shangabl, line 74: Not sure I like this test as there is nothing is added in the first place. Any string value would return nullptr.

test/util/TestFeatureRegistry.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+TEST_F(TestFeatureRegistry, StartAllFeatures)
+{
+    features->add(feature1->getName(), feature1);
+    features->add(feature2->getName(), feature2);
+    features->add(feature3->getName(), feature3);
+
+    features->startAll();
+
+    ASSERT_TRUE(feature1->isStarted());
```
shangabl, line 106: I think there is an ASSERT_ALL where you could group these together to be more readable.

source/FeatureRegistry.cpp
```diff
@@ -0,0 +1,78 @@
...
+        feature.second->stop();
+    }
+
+    for (auto &feature : features)
+    {
+        feature.second = nullptr;
+    }
+}
+
+void FeatureRegistry::startAll() const
```
RogerZhongAWS, line 69: you should check if the value is nullptr for startAll and stopAll

source/FeatureRegistry.cpp
```diff
@@ -0,0 +1,78 @@
...
+    {
+        features[name] = featureToAdd;
+    }
+    else
+    {
+        LOGM_WARN(TAG, "Attemped to add %s to Feature Registry despite it already existing", name.c_str());
+    }
+}
+
+void FeatureRegistry::remove(const std::string &name)
```
RogerZhongAWS, line 38: Since we now intend for nullptr to mean the feature is disabled, the meaning of the function name "remove" is no longer accurate. Consider renaming to "disable" or even "setNull"

source/FeatureRegistry.cpp
```diff
@@ -0,0 +1,78 @@
...
+    std::lock_guard<std::mutex> lock(featuresLock);
+    if (features.count(name))
+    {
+        return features.at(name);
+    }
+    LOGM_WARN(TAG, "Feature, %s, not found in current registry", name.c_str());
+    return nullptr;
+}
+
+void FeatureRegistry::add(std::string name, std::shared_ptr<Feature> featureToAdd)
```
RogerZhongAWS, line 25: It looks like we would need a "set" function to handle the cases where you enable a previously disabled feature. add() is only able to function upon startup since the features map hasn't been populated yet. We should discuss with Marco about whether we should create a new function or modify this one.

RogerZhongAWS (changes requested):

test/util/TestFeatureRegistry.cpp
adanforth: Not sure if I see it on here: https://google.github.io/googletest/reference/assertions.html

source/FeatureRegistry.cpp
adanforth: I think that from a users standpoint, it is essentially removed? They will be able to try and get the feature and we return that it is not there. There are no ways that we add more or less things into the registry after main.cpp runs. We simple have placeholder keys (this will be useful for dynamic reload) but for all intents and purposes, it is removed.

source/FeatureRegistry.cpp
adanforth: I think add serves a good purpose of literally adding features to the registry. Maybe a new method set() or something similar will be useful for dynamic reload. However, I will add this in the dynamic reload PR because it serves no purpose as of now.

source/FeatureRegistry.cpp
RogerZhongAWS: Yes, but the intended audience for the source code is developers, not users. If you have a class that is a wrapper around a container, its function names and behaviors are expected to be correspond with other examples of wrappers around containers.

shangabl (dismissed):

RogerZhongAWS (dismissed):

marcoemorais-aws (changes requested):

marcoemorais-aws (dismissed):

RogerZhongAWS (approved):

marcoemorais-aws (changes requested):
@adanforth 

## Issue
The normal shutdown path for device client will go through `shutdown()` function in `source/main.cpp`. In testing this change I discovered that the `shutdown()` function would hang waiting for resource cleanup as part of `aws_http_library_clean_up()`.

## Background
After digging a little deeper I learned the following things.

- The `aws_http_library_clean_up()` function is called from both the `~SecureTunnelingFeature()` destructor and the `~ApiHandle()` destructor in the c++ sdk. The ApiHandle is a member of SharedCrtResourceManager and used for communication with cloud via MQTT. By design, the lifetime of the ApiHandle and SharedCrtResourceManager are tied to the lifetime of the device client. As a result, we can remove the invocation of `aws_http_library_clean_up()`  from `~SecureTunnelingFeature()` destructor, since the cleanup function will be called as part of `~ApiHandle()` destructor.

```
-                SecureTunnelingFeature::~SecureTunnelingFeature() { aws_http_library_clean_up(); }
+                SecureTunnelingFeature::~SecureTunnelingFeature() = default;
```

- Prior to this change, I observed that there were 2 references to the SharedCrtResourceManager instance at the time of `shutdown()` (see output of `use_count()` below). This is unexpected and there should only be 1 reference after all features have stopped.  The extra reference prevented the destructor `~SharedCrtResourceManager()` from being invoked and as a result, the `~ApiHandle()` destructor was never invoked.

```
(gdb) p resourceManager
$1 = std::shared_ptr<Aws::Iot::DeviceClient::SharedCrtResourceManager> (use count 2, weak count 0) = {get() = 0x557c187a3330}
```

After this change, I observed exactly 1 reference to SharedCrtResourceManager after all features have stopped. 

```
Thread 1 "aws-iot-device-" hit Breakpoint 1, shutdown () at /src/source/main.cpp:150
150         resourceManager.reset();
(gdb) p resourceManager
$1 = std::shared_ptr<Aws::Iot::DeviceClient::SharedCrtResourceManager> (use count 1, weak count 0) = {get() = 0x55f6adb77bc0}
```

The single reference can now be released just prior to calling `exit()` and the `~ApiHandle()` destructor will be invoked. Please add the line `resourceManager.reset();` to the `shutdown()` function as shown below.

```
void shutdown()
{
    ...
    LoggerFactory::getLoggerInstance().get()->shutdown();
    resourceManager.reset();
    exit(0);
}
```

## Resolution
1. Modify `~SecureTunnelingFeature()` destructor to remove call to `aws_http_library_clean_up()`.
2. Modify `shutdown()` function to explicit release `SharedCrtResourceManager` just prior to `exit(0)` by calling `resourceManager.reset();`.

shangabl (approved):

marcoemorais-aws (approved):
👍

Issue #301, opened 2022-07-22
Incorrect permissions on created files when using fleet provisioning


JustinBerger:
**Describe the bug**

Changes permission of `/home/username/.aws-iot-device-client` directory to incorrect permissions (700 when it should be 745)
Creates runtime configuration file `/home/username/.aws-iot-device-client/aws-iot-device-client-runtime.conf` with incorrect permissions (644 when it should be 664)

**To Reproduce**

1. New project, build, setup.sh
2. Enable most resources, including provisioning
3. Cert & Key point to provisioning cert files
4. Do not install as service
5. Run ./aws-iot-device-client
6. File permission error with logs:
7. For sanity sake, set permissions correctly
chmod 745 /home/username/.aws-iot-device-client/
chmod 644 /home/username/.aws-iot-device-client/aws-iot-device-client-runtime.conf
8. re-run /aws-iot-device-client
9. still has permission error, and creates a new cert 

**Expected behavior**

Because `/home/username/.aws-iot-device-client/` already exists, it should not re-create it. 
If it creates it, it should be created with permission 745

`/home/username/.aws-iot-device-client/aws-iot-device-client-runtime.conf` should be created with 644

Should not crash with fatal error

**Actual behavior**

`/home/username/.aws-iot-device-client/` was changed to permission 700
`/home/username/.aws-iot-device-client/aws-iot-device-client-runtime.conf` was created with permission 664

**Logs**
Run 1
```
2022-07-22T19:04:19.789Z [WARN]  {Config.cpp}: Key {csr-file} was provided in the JSON configuration file with an empty value
2022-07-22T19:04:19.789Z [WARN]  {Config.cpp}: Key {device-key} was provided in the JSON configuration file with an empty value
2022-07-22T19:04:19.789Z [INFO]  {Config.cpp}: Successfully fetched JSON config file: 
    {
      "endpoint":       "my-endpoint",
      "cert":   "/aws-iot/keys/provisioning-certificate.pem.crt",
      "key":    "/aws-iot/keys/provisioning-private.pem.key",
      "root-ca":        "/aws-iot/keys/AmazonRootCA1.pem",
      "thing-name":     "my_device_name",
      "logging":        {
        "level":        "DEBUG",
        "type": "STDOUT",
        "file": "/var/log/aws-iot-device-client/aws-iot-device-client.log",
        "enable-sdk-logging":   false,
        "sdk-log-level":        "TRACE",
[...]

PR #305, opened 2022-07-27
Removed code which was redundantly creating  dir with incorrect permissions. 

HarshGandhi-AWS:
### Motivation
- Issue number: #301 
- We were creating directory files with incorrect file permissions. 

#### Change summary

#### Updated code to:
* Not to redundantly create default files with user has already created and passed file path. 
* Not to create file with incorrect permissions 

#### Revision diff summary

### Testing
- CI test run result: <link>
- Manually tested the changes.

### DIFF
--- a/source/config/Config.cpp
+++ b/source/config/Config.cpp
@@ -1586,13 +1586,6 @@ bool PlainConfig::SampleShadow::LoadFromJson(const Crt::JsonView &json)
         {
             shadowOutputFile = FileUtils::ExtractExpandedPath(json.GetString(jsonKey).c_str());
         }
-        else
-        {
-            if (!createShadowOutputFile())
-            {
-                return false;
-            }
-        }
     }
 
     return true;
@@ -1619,7 +1612,9 @@ bool PlainConfig::SampleShadow::LoadFromCliArgs(const CliArgs &cliArgs)
             FileUtils::ExtractExpandedPath(cliArgs.at(PlainConfig::SampleShadow::CLI_SAMPLE_SHADOW_OUTPUT_FILE))
                 .c_str();
     }
-    else
+
+    // setting `shadowOutputFile` value to default if no value was passed by user via CLI or JSON config.
+    if (!shadowOutputFile.has_value() || shadowOutputFile->empty())
     {
         if (!createShadowOutputFile())
         {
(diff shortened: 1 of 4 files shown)


source/config/Config.cpp
```diff
@@ -1657,6 +1652,11 @@ bool PlainConfig::SampleShadow::Validate() const
         }
         else
         {
+            LOGM_ERROR(
+                Config::TAG,
+                "*** %s: Invalid file path passed for argument: %s ***",
+                DeviceClient::DC_FATAL_ERROR,
+                JSON_SAMPLE_SHADOW_INPUT_FILE);
```
marcoemorais-aws, line 1659: You could print the value of the `shadowInputFile` to help the user more quickly identify the problem. Same goes for `shadowOutputFile`.

source/fleetprovisioning/FleetProvisioning.cpp
```diff
@@ -494,7 +494,8 @@ bool FleetProvisioning::ProvisionDevice(shared_ptr<SharedCrtResourceManager> fpC
         LOG_INFO(TAG, "Fleet Provisioning Feature has been started.");
 
         bool didSetup = FileUtils::CreateDirectoryWithPermissions(keyDir.c_str(), S_IRWXU) &&
-                        FileUtils::CreateDirectoryWithPermissions(Config::DEFAULT_CONFIG_DIR, S_IRWXU);
+                        FileUtils::CreateDirectoryWithPermissions(
+                            Config::DEFAULT_CONFIG_DIR, S_IRUSR | S_IWUSR | S_IXUSR | S_IRGRP | S_IROTH | S_IXOTH);
```
marcoemorais-aws, line 498: Isn't this the same as `Config::CONFIG_DIR` eg (octal 745 / `rwxr--r-x`). If so, then maybe better to reference the constant.

source/config/Config.cpp
HarshGandhi-AWS: updated the log statements.

source/fleetprovisioning/FleetProvisioning.cpp
HarshGandhi-AWS: Since this change needs to be applied at multiple places, I will have a separate PR for updating this and adding unit tests for the same.

xlcheng1 (approved):

RogerZhongAWS (approved):

Issue #282, opened 2022-06-09
JobEngine not processing Job Documents as expected


climbertjh2:
I have been trying to use the *Job Document* features to submit jobs to a test device running the `aws-iot-device-client` on Linux Ubuntu.

I have found that the parameters passed in the `args` array of the Job Document as well as the `runAsUser`, if specified, are not being handled or processed as expected.

Here are two examples of what I have found to work vs. not work, along with the corresponding `aws-iot-device-client.log` output:

**failing example**

Script
```
#!/bin/bash

name="example-job-fails-02"
thing="arn:aws:iot:us-east-1:NNNNNNNNNNNN:thing/deviceClientThing"
document="{ \"operation\": \"sh\", \"args\": [ \"-c\", \"'echo hello'\" ] }"

aws iot create-job \
    --job-id "$name" \
    --targets "$thing" \
    --document "$document" \
    --description "example job test - this job fails" \
    --target-selection SNAPSHOT
```
Log output:
```
2-06-09T14:46:11.058Z [INFO]  {JobsFeature.cpp}: Executing job: example-job-fails-02
2022-06-09T14:46:11.058Z [DEBUG] {Retry.cpp}: Retryable function starting, it will retry until success
2022-06-09T14:46:11.058Z [DEBUG] {JobsFeature.cpp}: Created EphemeralPromise for ClientToken 0pn5h7pCCm in the updateJobExecution promises map
2022-06-09T14:46:11.058Z [INFO]  {JobEngine.cpp}: About to execute step with name: sh
2022-06-09T14:46:11.058Z [DEBUG] {JobEngine.cpp}: Assuming executable is in PATH
2022-06-09T14:46:11.058Z [INFO]  {JobEngine.cpp}: About to execute: sh  -c 'echo hello'
2022-06-09T14:46:11.059Z [DEBUG] {JobEngine.cpp}: Child process now running
2022-06-09T14:46:11.059Z [DEBUG] {JobEngine.cpp}: Child process about to call execvp
2022-06-09T14:46:11.059Z [DEBUG] {JobEngine.cpp}: Parent process now running, child PID is 1006
2022-06-09T14:46:11.059Z [ERROR] {1006}: sh: 0: Can't open
2022-06-09T14:46:11.059Z [DEBUG] {JobEngine.cpp}: JobEngine finished waiting for child process, returning 32512
[...]

PR #306, opened 2022-07-28
wrapping shell-command job into shell-command-handler.sh

xlcheng1:
### Motivation
- JobEngine is not supporting running shell-command jobs, this change provides a work around by wrapping shell-command job using one of the sample-job-handler: shell-command-handler.sh
- Issue number: #282 

#### Change summary
Invoking shell-command-handler.sh when action.name = "shell-command"

#### Revision diff summary

### Testing
- CI test run result: <link>

### DIFF
--- a/source/jobs/JobEngine.cpp
+++ b/source/jobs/JobEngine.cpp
@@ -139,21 +139,20 @@ void JobEngine::exec_action(PlainJobDocument::JobAction action, const std::strin
     string command;
     if (action.type == PlainJobDocument::ACTION_TYPE_RUN_HANDLER)
     {
-        try{
+        try
+        {
             // build command for shell command
             if (action.name == "shell-command")
             {
                 command = buildCommand(jobHandlerDir, "shell-command-handler.sh", jobHandlerDir);
-
             }
             else
             {
                 // build command for standard jobHandler
                 command = buildCommand(action.input.path, action.input.handler, jobHandlerDir);
-
             }
-
-        }catch (exception &e)
+        }
+        catch (exception &e)
         {
             if (!action.ignoreStepFailure.value())
             {
--- a/source/jobs/JobEngine.cpp
+++ b/source/jobs/JobEngine.cpp
@@ -141,16 +141,24 @@ void JobEngine::exec_action(PlainJobDocument::JobAction action, const std::strin
     {
         try
         {
+            // build command for standard jobHandler
+            command = buildCommand(action.input.path, action.input.handler, jobHandlerDir);
+        }
+        catch (exception &e)
+        {
+            if (!action.ignoreStepFailure.value())
+            {
+                executionStatus = 1;
+            }
+            return;
+        }
+    }
+    else if (action.type == "runShellCommand")
+    {
+        try
+        {
             // build command for shell command
-            if (action.name == "shell-command")
-            {
-                command = buildCommand(jobHandlerDir, "shell-command-handler.sh", jobHandlerDir);
-            }
-            else
-            {
-                // build command for standard jobHandler
-                command = buildCommand(action.input.path, action.input.handler, jobHandlerDir);
-            }
+            command = buildCommand(jobHandlerDir, "shell-command-handler.sh", jobHandlerDir);
         }
         catch (exception &e)
         {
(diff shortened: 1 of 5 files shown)


source/jobs/JobEngine.cpp
```diff
@@ -139,18 +139,37 @@ void JobEngine::exec_action(PlainJobDocument::JobAction action, const std::strin
     string command;
     if (action.type == PlainJobDocument::ACTION_TYPE_RUN_HANDLER)
     {
-        // build command
-        try
+        // build command for shell command
+        if (action.name == "shell-command")
```
marcoemorais-aws, line 143: Nit. You can prolly put the try-catch around the `if-else` since the catch handler is the same.

marcoemorais-aws (changes requested):
We don't have any documentation for `action.name=shell-command`. Please add details about how this action works and include an example with a job document that should be passed to the jobs api.

https://github.com/awslabs/aws-iot-device-client/blob/main/source/jobs/README.md

source/jobs/JobEngine.cpp
xlcheng1: addressed

source/jobs/JobEngine.cpp
```diff
@@ -153,6 +153,22 @@ void JobEngine::exec_action(PlainJobDocument::JobAction action, const std::strin
             return;
         }
     }
+    else if (action.type == "runShellCommand")
```
HarshGandhi-AWS, line 156: why are we having separate block for building command for `shell-command-handler`?  I think without changing anything we can still execute the same handler file. And anyways the handler file name is passed in job document.

source/jobs/JobEngine.cpp
```diff
@@ -153,6 +153,22 @@ void JobEngine::exec_action(PlainJobDocument::JobAction action, const std::strin
...
+    {
+        try
+        {
+            // build command for shell command
+            command = buildCommand(jobHandlerDir, "shell-command-handler.sh", jobHandlerDir);
```
HarshGandhi-AWS, line 161: Another issue I see is we are hard coding the handler name here. This means user cannot have (1)handler with different name (2)multiple shell command handler.

HarshGandhi-AWS (changes requested):

source/jobs/JobEngine.cpp
xlcheng1: it is not an issue because shell-command-handler.sh is treated as a standard job handler and is created by us under sample-job-handler directory, whatever the shell-command job document will be, device client will invoke this handler to process

source/jobs/JobEngine.cpp
xlcheng1: this is needed because some customers e.g. the customer who reported the ticket, he puts `/bin` as the path, this will cause device client not being able to find `shell-command-handler.sh` under `sample-job-handler` directory

HarshGandhi-AWS: Hey @marcoemorais-aws , what are your thought on my this and other comment below? I think if user is passing a different directory path, we should execute the handler present in that directory.

RogerZhongAWS: It does seem odd that we do not honor the input path for this runShellCommand type. I think we need to look at this again and make a decision for whether we really need this unique call to buildCommand

shangabl: One thing we discussed before was just having a handler that just builds the command and invokes the next handler... We could just pass all runHandler jobs to this preHandler and it check user and format the command and then invoke the handler in the path.

BolongZhang-AWS (commented):
Have we reached any conclusion regarding the file path/handler input issue?

Other than that I only some minor comments regarding the format: 
1. When hard coding strings or numbers let's define constants in `*.c` files then use them in the logic.
2. For the example can we also give an example to echo hello world? I think previously there's a customer having problem with the quotes.

xlcheng1:
> Have we reached any conclusion regarding the file path/handler input issue?
> 
> Other than that I only some minor comments regarding the format:
> 
>     1. When hard coding strings or numbers let's define constants in `*.c` files then use them in the logic.
> 
>     2. For the example can we also give an example to echo hello world? I think previously there's a customer having problem with the quotes.

We will not use this PR as solution, we are going to use c++ code to replace the wrapper script to handle a seperate `type`: `runCommand` and the structure of `runCommand` will be different, it will not have `handler` and `path`, WIP

xlcheng1:
no longer needed, RunCommand is implemented, see #359

PR #314, opened 2022-08-11
WIP config serialization 

adanforth:
### Motivation
Adding more Serialization methods to config subclasses in order to export a config in its entirety.

#### Change summary
- Added serialization and helper functions to Config class
- Created equality logic for device defender - probably should keep this dynamic reload PR.
- Added two basic tests

### Testing
 - Simple unit tests. TODO


test/config/TestConfig.cpp
```diff
@@ -1026,6 +1099,86 @@ TEST_F(ConfigTestFixture, DeviceDefenderCli)
     ASSERT_EQ(6, config.deviceDefender.interval);
 }
 
+TEST_F(ConfigTestFixture, DeviceDefenderComparison)
```
shangabl, line 1102: Seems like 3 of these configs are exactly the same? It isn't necessary to test like that. That would be like testing the code to see if its non-deterministic. Create a helper method that allows you to input some parameters to output a jsonString and test with different parameters and assert they are equal.

test/config/TestConfig.cpp
```diff
@@ -231,6 +231,79 @@ TEST_F(ConfigTestFixture, AllFeaturesEnabled)
...
+
+    JsonObject serializedConfig;
+    config.SerializeToObject(serializedConfig);
+
+    jsonView = serializedConfig.View();
+
+    PlainConfig postSerializedConfig;
+    postSerializedConfig.LoadFromJson(jsonView);
+
+    ASSERT_EQ(config.endpoint.value(), postSerializedConfig.endpoint.value());
```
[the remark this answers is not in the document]
marcoemorais-aws: From offline discussion with Andy, he can shortcut the tedious attribute-by-attribute assertion by comparing minified json strings of the serialized config with the postSerializedConfig. `JsonView` class has a method `WriteCompact` to generate the minified json string.

https://github.com/awslabs/aws-crt-cpp/blob/c95c5699d0f8da9b3fa402fc41e163248efb3ba6/include/aws/crt/JsonObject.h#L383-L387

source/config/Config.h
```diff
@@ -69,6 +69,8 @@ namespace Aws
                 bool LoadFromCliArgs(const CliArgs &cliArgs) override;
                 bool LoadFromEnvironment() override;
                 bool Validate() const override;
+                /** Serialize configurations To Json Object **/
+                void SerializeToObject(Crt::JsonObject &object) const;
```
marcoemorais-aws, line 73: Parameter should be const.

```
void SerializeToObject(const Crt::JsonObject &object) const;
```

marcoemorais-aws (changes requested):

source/config/Config.h
```diff
@@ -200,6 +204,8 @@ namespace Aws
                     bool LoadFromCliArgs(const CliArgs &cliArgs) override;
                     bool LoadFromEnvironment() override { return true; }
                     bool Validate() const override;
+                    bool operator==(const PlainConfig::DeviceDefender &other) const;
```
marcoemorais-aws, line 207: Don't include this as part of this PR.

test/config/TestConfig.cpp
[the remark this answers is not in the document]
marcoemorais-aws: You can turn this into a Value-Parameterized tests [1] by passing different `jsonString` to the block of code below. 

```
    PlainConfig config;
    JsonObject jsonObject(jsonString);
    config.LoadFromJson(jsonObject.View());
    auto inputJsonString = jsonObject.View().WriteCompact();

    JsonObject serializedConfig;
    config.SerializeToObject(serializedConfig);
    auto serializedJsonString = serializedConfig.View().WriteCompact();

    ASSERT_STREQ(inputJsonString.c_str(), serializedJsonString.c_str());
```

When I run this block for the all-features-enabled-string, I see some errors relating to how we are serializing the log level.

```
Expected equality of these values:
  inputJsonString.c_str()
    Which is: "{\"endpoint\":\"endpoint value\",\"cert\":\"/tmp/aws-iot-device-client-test-file\",\"key\":\"/tmp/aws-iot-device-client-test-file\",\"thing-name\":\"thing-name value\",\"logging\":{\"level\":\"debug\",\"type\":\"file\",\"file\":\"./aws-iot-device-client.log\"},\"jobs\":{\"enabled\":true},\"tunneling\":{\"enabled\":true},\"device-defender\":{\"enabled\":true,\"interval\":300},\"fleet-provisioning\":{\"enabled\":true,\"template-name\":\"template-name\",\"csr-file\":\"/tmp/aws-iot-device-client-test-file\",\"device-key\":\"/tmp/aws-iot-device-client-test-file\",\"template-parameters\":\"{\\\"SerialNumber\\\": \\\"Device-SN\\\"}\"},\"samples\":{\"pub-sub\":{\"enabled\":true,\"publish-topic\":\"publish_topic\",\"subscribe-topic\":\"subscribe_topic\"}},\"config-shadow\":{\"enabled\":true},\"sample-shadow\":{\"enabled\":true,\"shadow-name\":\"shadow-name\",\"shadow-input-file\":\"\",\"shadow-output-file\":\"\"},\"secure-element\":{\"enabled\":true,\"pkcs11-lib\":\"/tmp/aws-iot-device-client-test-file\",\"secure-element-pin\":\"0000\",\"secure-element-key-label\":\"key-label\",\"secure-element-slot-id\":1111,\"secure-element-token-label\":\"token-label\"}}"
  serializedJsonString.c_str()
    Which is: "{\"endpoint\":\"endpoint value\",\"cert\":\"/tmp/aws-iot-device-client-test-file\",\"key\":\"/tmp/aws-iot-device-client-test-file\",\"thing-name\":\"thing-name value\",\"logging\":{\"level\":3,\"type\":\"file\",\"file\":\"./aws-iot-device-client.log\",\"enable-sdk-logging\":false,\"sdk-log-level\":\"TRACE\",\"sdk-log-file\":\"/var/log/aws-iot-device-client/sdk.log\"},\"jobs\":{\"enabled\":true,\"handler-directory\":\"\"},\"tunneling\":{\"enabled\":true},\"device-defender\":{\"enabled\":true,\"interval\":300},\"fleet-provisioning\":{\"enabled\":true,\"template-name\":\"template-name\",\"template-parameters\":\"{\\\"SerialNumber\\\": \\\"Device-SN\\\"}\",\"csr-file\":\"/tmp/aws-iot-device-client-test-file\",\"device-key\":\"/tmp/aws-iot-device-client-test-file\"},\"runtime-config\":{\"completed-fp\":true},\"samples\":{\"pub-sub\":{\"enabled\":true,\"publish-topic\":\"publish_topic\",\"subscribe-topic\":\"subscribe_topic\"}},\"config-shadow\":{\"enabled\":true},\"sample-shadow\":{\"enabled\":true,\"shadow-name\":\"shadow-name\"},\"secure-element\":{\"enabled\":true,\"pkcs11-lib\":\"/tmp/aws-iot-device-client-test-file\",\"secure-element-pin\":\"0000\",\"secure-element-key-label\":\"key-label\",\"secure-element-slot-id\":1111,\"secure-element-token-label\":\"token-label\"}}"
```

[1] https://google.github.io/googletest/advanced.html#value-parameterized-tests

source/config/Config.h
```diff
@@ -210,6 +215,9 @@ namespace Aws
                     static constexpr char JSON_KEY_INTERVAL[] = "interval";
 
                     bool enabled{false};
+                    // userDefined keeps track of whether or not the feature was defined by a user when loaded from a
+                    // json. This is only to be used by dynamic reload.
+                    bool userDefined{false};
```
marcoemorais-aws, line 220: Prolly want to keep this out of this review.

source/config/Config.cpp
```diff
@@ -539,6 +615,32 @@ string PlainConfig::LogConfig::ParseDeviceClientLogType(string value)
     }
 }
 
+string PlainConfig::LogConfig::StringifyDeviceClientLogLevel(int level) const
+{
+
+    switch (level)
+    {
+        case 0:
+            return "error";
```
marcoemorais-aws, line 624: Should probably be ALL CAPS to be consistent with `PlainConfig::LogConfig::ParseDeviceClientLogLevel`.

source/config/Config.cpp
```diff
@@ -539,6 +615,32 @@ string PlainConfig::LogConfig::ParseDeviceClientLogType(string value)
...
+            break;
+        case 1:
+            return "warn";
+            break;
+        case 2:
+            return "info";
+            break;
+    }
+    // this is the defualt case for log level
```
marcoemorais-aws, line 633: * Add an explicit case for Debug to be consistent with out-of-range handling in the other functions that convert log level.
* Replace the magic numbers with the named constants. This requires you to use a `static_cast` to convert `int` to `enum class LogLevel`.

```
switch (static_cast<DeviceClient::Logging::LogLevel>(level)):
{
    case DeviceClient::Logging::LogLevel::ERROR:
        return "ERROR";
    ...
    case DeviceClient::Logging::LogLevel::DEBUG:
        return "DEBUG";
}
throw std::invalid_argument(...);
```

source/config/Config.cpp
```diff
@@ -433,6 +433,82 @@ bool PlainConfig::Validate() const
...
+
+    JsonObject deviceDefenderObject;
+    deviceDefender.SerializeToObject(deviceDefenderObject);
+    object.WithObject(JSON_KEY_DEVICE_DEFENDER, deviceDefenderObject);
+
+    JsonObject fleetProvisioningObject;
+    fleetProvisioning.SerializeToObject(fleetProvisioningObject);
+    object.WithObject(JSON_KEY_FLEET_PROVISIONING, fleetProvisioningObject);
+
+    if (fleetProvisioning.enabled)
```
marcoemorais-aws, line 479: Since we said we would serialize features that aren't enabled, doesn't that mean we need to remove this check?

source/config/Config.cpp
```diff
@@ -433,6 +433,82 @@ bool PlainConfig::Validate() const
...
+
+    JsonObject sampleShadowObject;
+    sampleShadow.SampleShadow::SerializeToObject(sampleShadowObject);
+    object.WithObject(JSON_KEY_SAMPLE_SHADOW, sampleShadowObject);
+
+    JsonObject secureElementObject;
+    secureElement.SerializeToObject(secureElementObject);
+    object.WithObject(JSON_KEY_SECURE_ELEMENT, secureElementObject);
+
+    if (sensorPublish.enabled)
```
marcoemorais-aws, line 504: Same as above, should we remove this check?

source/config/Config.cpp
```diff
@@ -2248,6 +2405,76 @@ bool PlainConfig::SensorPublish::Validate() const
...
+        }
+
+        if (entry.heartbeatTimeSec.has_value())
+        {
+            sensor.WithInt64(JSON_HEARTBEAT_TIME_SEC, entry.heartbeatTimeSec.value());
+        }
+
+        sensors.push_back(sensor);
+    }
+
```
marcoemorais-aws, line 2474: I think you are missing the enabled flag for the feature itself.

marcoemorais-aws (changes requested):

source/config/Config.h
adanforth: Agreed.

source/config/Config.cpp
adanforth: Yes, I agree, I was checking out the other configs defined in Config.cpp and they used this lower case version, but it SHOULD be upper.

source/config/Config.cpp
adanforth: yup, this was left in from earlier

source/config/Config.cpp
adanforth: I want to keep this one since otherwise they have to define an empty array of sensors. But this is easily removed and the tests can be fixed.

source/config/Config.cpp
adanforth: Resolved offline, will keep the check

marcoemorais-aws (approved):
👍

shangabl (approved):

PR #321, opened 2022-09-07
Sonarcloud static analysis

RogerZhongAWS:
### Motivation
- Enables sonarqube to analyze our code for bugs, security, and style issues
#### Change summary
- Fixes for most of the issues identified by sonarqube
- Autoformatting changes

#### Revision diff summary

### Testing
- CI test run result: <link>


BolongZhang-AWS (dismissed):
Can we add some documentation for how to navigate and use Sonarcloud?

.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,47 @@
...
+    branches:
+      - '*'
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  sonarcloud:
+    runs-on: ubuntu-latest
+    if: (github.event_name == 'push') || ((github.event_name == 'pull_request') && (github.event.pull_request.head.repo.full_name != github.repository))
```
shangabl, line 13: The first two conditionals are redundant I think?

shangabl (commented):
Can this be run locally? It would be good to have a script like we do for checking format and cppcheck so when we are developing we can just mount to a docker container and run the scripts.

.github/workflows/sonarcloud.yml
RogerZhongAWS: they are needed to prevent duplicate runs from triggering if a new update is made to a PR

RogerZhongAWS:
> Can this be run locally? It would be good to have a script like we do for checking format and cppcheck so when we are developing we can just mount to a docker container and run the scripts.

yes, its pretty straightforward.

This PR pretty much already contains the script for doing so. You need to:
1. Add the compile flag that disables optimization "-O0" and enable -fprofile-arcs -ftest-coverage which generates control flow information
2. Run the tests as you normally would.
3. Find the directory where the .gcda files are stored. This is by default the same directory that the object files *.o are written to
4. Run the gcov command with a file name or for a directory, use the same command in the sonarcloud.yml file to convert the .gcda into a human readable output.
5. Add whatever other steps you want to collate the data.

shangabl:
> > Can this be run locally? It would be good to have a script like we do for checking format and cppcheck so when we are developing we can just mount to a docker container and run the scripts.
> 
> yes, its pretty straightforward.
> 
> This PR pretty much already contains the script for doing so. You need to:
> 
> 1. Add the compile flag that disables optimization "-O0" and enable -fprofile-arcs -ftest-coverage which generates control flow information in the form of a .gcno file
> 2. Run the tests as you normally would. This will write encoded coverage information to .gcda files
> 3. Find the directory where the .gcda files are stored. This is by default the same directory that the object files *.o are written to
> 4. Run the gcov command with a file name or for a directory, use the same command in the sonarcloud.yml file to convert the .gcda into a human readable output.
> 5. Add whatever other steps you want to collate the data.

I think it would be good to have a bash script for this

shangabl (approved):

xlcheng1 (approved):

PR #327, opened 2022-09-13
Fix Fleet Provisioning Documentation

shangabl:
### Motivation
FP Documentation is incorrect. The provided example Claim Certificate Policy is incorrect and doesn't work. Tested manually, tested using the provided example here https://docs.aws.amazon.com/iot/latest/developerguide/provision-wo-cert.html which was successful.

### Testing
Manually provisioned device using containers.


HarshGandhi-AWS (approved):
LGTM

PR #332, opened 2022-09-22
Update lockfile handler

RogerZhongAWS:
### Motivation
- Changes lockfile behavior in accordance to usability needs.

#### Change summary
- now writes thing name to lockfile

### DIFF
--- a/integration-tests/entry-point.sh
+++ b/integration-tests/entry-point.sh
@@ -59,20 +59,21 @@ service ssh start || true
 
 # Ensure lockfile exists
 mkdir -p /run/lock > dev/null
-touch /run/lock/devicecl.lock
 
 # Spoof lockfile
 echo "${THING_NAME}" > /run/lock/devicecl.lock
-bash -c 'exec -a aws-iot-device-client sleep 1000000' &
+bash -c 'exec -a aws-iot-device-client sleep 1000' &
 PID=$!
 echo ${PID} >> /run/lock/devicecl.lock
 
 # Start and stop Device Client
-./aws-iot-device-client 2>&1 &
-sleep 2
-pkill -f aws-iot-device-client
+./aws-iot-device-client
+retVal=$?
+if [ $retVal -ne 1 ]; then
+    exit 1
+fi
 
-# cleanup
+# Cleanup
 rm -rf /run/lock
 
 # start Device Client
--- a/integration-tests/entry-point.sh
+++ b/integration-tests/entry-point.sh
@@ -63,19 +63,20 @@ mkdir -p /run/lock > dev/null
 # Spoof lockfile
 echo "${THING_NAME}" > /run/lock/devicecl.lock
 bash -c 'exec -a aws-iot-device-client sleep 1000' &
-PID=$!
-echo ${PID} >> /run/lock/devicecl.lock
+FAKE_PID=$!
+echo ${FAKE_PID} >> /run/lock/devicecl.lock
 
 # TEST: Start and stop Device Client
 # 1. Run DC
 # 2. Kill dummy process
 # 3. Kill Device Client. If Device Client has already exited as expected, then pkill will return 1, and we pass the test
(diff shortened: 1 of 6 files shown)


shangabl:
Can you add an integration test for this? Or at least remove the part of the integration tests entry-point.sh script that create the /run/lock directory and make sure it still works. We may want to add a test though where we try to run DC again and verify it fails.

integration-tests/entry-point.sh
```diff
@@ -60,6 +57,16 @@ chmod 640 ${CONFIG_PATH}
 # start ssh
 service ssh start || true
 
+# Start Device Client for a short time without lockfile dir
+rm -rf /run/lock
+./aws-iot-device-client 2>&1 &
+sleep 2
```
shangabl, line 63: This actually makes the code below which waits for the runtime config file useless.

Also, /run/lock doesn't exist already so it isn't really testing anything.

Maybe create the lock file, verify DC fails to start, then delete it?

shangabl (changes requested):

integration-tests/entry-point.sh
RogerZhongAWS: i see, creating the lockfile will be tricky without starting the device client first since it needs a valid PID to be written into it

integration-tests/entry-point.sh
```diff
@@ -60,6 +57,24 @@ chmod 640 ${CONFIG_PATH}
 # start ssh
 service ssh start || true
 
+# Ensure lockfile exists
+mkdir -p /run/lock > dev/null
+touch /run/lock/devicecl.lock
```
shangabl, line 62: This is redundant because > and >> will both create the file

integration-tests/entry-point.sh
```diff
@@ -60,6 +57,24 @@ chmod 640 ${CONFIG_PATH}
...
+
+# Spoof lockfile
+echo "${THING_NAME}" > /run/lock/devicecl.lock
```
shangabl, line 65: I'm confused about what exactly is supposed to happen here.

integration-tests/entry-point.sh
shangabl: I think instead of sleeping here you should be verifying that the DC exited?

integration-tests/entry-point.sh
RogerZhongAWS: >> appends to the end without overwriting the beginning

integration-tests/entry-point.sh
RogerZhongAWS: We now check for thing name as well as pid since we may want to support the use case of running multiple DC instances with different thing names

integration-tests/entry-point.sh
```diff
@@ -60,6 +57,25 @@ chmod 640 ${CONFIG_PATH}
...
+bash -c 'exec -a aws-iot-device-client sleep 1000' &
+PID=$!
+echo ${PID} >> /run/lock/devicecl.lock
+
+# Start and stop Device Client
+./aws-iot-device-client
+retVal=$?
+if [ $retVal -ne 1 ]; then
+    exit 1
```
shangabl, line 73: I would add more comments walking through this... This is actually an integration test but it doesn't say that.

Lets explicitly call this out as as a test, explain each step. 

And can you add some stdout for the failure case? If our integration tests just stop running because of a lock file issue we want to know why

RogerZhongAWS: Modified the test to no longer check for the return code of device client (in the case where device client keeps running and does not exit) but instead checks for the return code of pkill after we kill the dummy process first. This value will be 1 if there is no such process named 'aws-iot-device-client'

shangabl (dismissed):

RogerZhongAWS:
note: add exception handler for stoi call

shangabl (approved):

xlcheng1 (approved):

PR #368, opened 2022-12-21
Add Device Defender Integration Tests

shangabl:
### Motivation
- Need integration tests for Device Defender

#### Change summary
Added Integration tests for DD
Fixed typo in ubi8 docker image tag
Added a Download File integration test for Jobs

#### Revision diff summary

### Testing
- CI test run result: <link>

### DIFF
--- a/integration-tests/source/IntegrationTestResourceHandler.cpp
+++ b/integration-tests/source/IntegrationTestResourceHandler.cpp
@@ -73,7 +73,7 @@ void IntegrationTestResourceHandler::CreateJob(const string &jobId, const string
     }
     else
     {
-        LogError("Failed to Create Job", jobId, outcome.GetError().GetMessage());
+        Log(Logging::LogLevel::Error, "Failed to Create Job", jobId, outcome.GetError().GetMessage());
     }
 }
 
@@ -87,7 +87,7 @@ JobExecutionStatus IntegrationTestResourceHandler::GetJobExecutionStatus(const s
 
     if (!outcome.IsSuccess())
     {
-        LogError("Failed to describe job execution for Job", jobId, outcome.GetError().GetMessage());
+        Log(Logging::LogLevel::Error, "Failed to describe job execution for Job", jobId, outcome.GetError().GetMessage());
     }
     return outcome.GetResult().GetExecution().GetStatus();
 }
@@ -100,7 +100,7 @@ void IntegrationTestResourceHandler::DeleteJob(const std::string &jobId)
     DeleteJobOutcome outcome = iotClient.DeleteJob(request);
     if (!outcome.IsSuccess())
     {
-        LogError("Failed to Delete Job", jobId, outcome.GetError().GetMessage());
+        Log(Logging::LogLevel::Error, "Failed to Delete Job", jobId, outcome.GetError().GetMessage());
     }
 }
 void IntegrationTestResourceHandler::DeleteThing(const std::string &thingName)
@@ -112,7 +112,7 @@ void IntegrationTestResourceHandler::DeleteThing(const std::string &thingName)
 
     if (!outcome.IsSuccess())
     {
-        LogError("Failed to delete Thing", thingName, outcome.GetError().GetMessage());
+        Log(Logging::LogLevel::Error, "Failed to delete Thing", thingName, outcome.GetError().GetMessage());
     }
 }
 
@@ -132,7 +132,7 @@ vector<Aws::Utils::ARN> IntegrationTestResourceHandler::ListCertsForThing(const 
     }
     else
     {
-        LogError("Failed to list Certificates for Thing", thingName, outcome.GetError().GetMessage());
+        Log(Logging::LogLevel::Error, "Failed to list Certificates for Thing", thingName, outcome.GetError().GetMessage());
     }
 
     return certs;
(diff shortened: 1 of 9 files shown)


integration-tests/source/devicedefender/DeviceDefenderIntegrationTests.cpp
```diff
@@ -0,0 +1,73 @@
...
+        Aws::ShutdownAPI(options);
+    }
+    unique_ptr<IntegrationTestResourceHandler> resourceHandler;
+    vector<std::string> metrics{"aws:all-bytes-in", "aws:all-bytes-out", "aws:all-packets-in", "aws:all-packets-out"};
+};
+
+/**
+ * To test Device Defender we are creating a Security Profile to create violations for any metrics output and using this
+ * to facilitate testing this is a simple test to verify the metrics are being emitted by the Device Client by verifying
+ * all metrics are causing violations
```
RogerZhongAWS, line 49: Just to be clear we are just trivially checking in bytes/packets in-out is less than 1 which should trigger a violation?

integration-tests/source/IntegrationTestResourceHandler.cpp
```diff
@@ -323,3 +329,102 @@ std::string IntegrationTestResourceHandler::GetResourceId(const std::string &res
...
+{
+    string allThingsArn = targetArn.substr(0, targetArn.find_last_of(':') + 1) + "all/things";
+
+    AttachSecurityProfileRequest request;
+    request.SetSecurityProfileName(profileName);
+    request.SetSecurityProfileTargetArn(allThingsArn);
+
+    AttachSecurityProfileOutcome outcome = iotClient.AttachSecurityProfile(request);
+
+    if (!outcome.IsSuccess())
```
RogerZhongAWS, line 404: duplicate conditional

integration-tests/source/IntegrationTestResourceHandler.cpp
```diff
@@ -323,3 +329,102 @@ std::string IntegrationTestResourceHandler::GetResourceId(const std::string &res
...
+        printf(
+            "Failed to create Security Profile: %s\n%s\n",
+            profileName.c_str(),
+            outcome.GetError().GetMessage().c_str());
+    }
+
+    AttachSecurityProfile(profileName);
+}
+
+void IntegrationTestResourceHandler::AttachSecurityProfile(const std::string &profileName)
```
RogerZhongAWS, line 394: Curious to know whether we will ever call this method on its own (not as a part of CreateAndAttachSecurityProfile). I think we either make this method private or remove the call from line 391 and call this directly from test setup.

shangabl: I think I meant to make it private but made it protected by mistake. Will change.

integration-tests/source/devicedefender/DeviceDefenderIntegrationTests.cpp
shangabl: Yes I can update that to make it clearer.

This was the only way I could think to test. DD has Device Side Metrics and Cloud Side Metrics and of course we only want to test the device and don't support custom metrics. So, verifying that we are in fact emitting these metrics seemed to be the best way.

integration-tests/source/IntegrationTestResourceHandler.cpp
```diff
@@ -202,6 +219,7 @@ void IntegrationTestResourceHandler::CleanUpThingAndCert(const std::string &thin
         DeleteCertificate(certificateId);
         DeleteThing(thingName);
     }
+    DeleteThingGroup(thingName);
```
RogerZhongAWS, line 222: is this intentional

RogerZhongAWS (dismissed):

integration-tests/source/IntegrationTestResourceHandler.cpp
shangabl: Nope... Thanks for catching this.

RogerZhongAWS (approved):

xlcheng1 (approved):

integration-tests/source/IntegrationTestResourceHandler.cpp
```diff
@@ -323,3 +342,155 @@ std::string IntegrationTestResourceHandler::GetResourceId(const std::string &res
...
+{
+    MetricValue metricValue;
+    metricValue.SetCount(1L);
+
+    BehaviorCriteria criteria;
+    criteria.SetComparisonOperator(ComparisonOperator::less_than);
+    criteria.SetDurationSeconds(300);
+    criteria.SetValue(metricValue);
+    criteria.SetConsecutiveDatapointsToAlarm(1);
+    criteria.SetConsecutiveDatapointsToClear(1);
```
RogerZhongAWS, line 384: do you think this line can be causing any issue? would probably be safer to set this to 100 or something

shangabl: I hope that is the issue. Just changed! Thanks!

xlcheng1 (dismissed):

RogerZhongAWS (approved):

xlcheng1 (approved):

RogerZhongAWS (approved):

xlcheng1 (approved):

PR #391, opened 2023-03-03
Patch for Greengrass Secure Tunneling failure to start

RogerZhongAWS:
### Motivation
- Moving the call to lockfile handler to indef, so it works with greengrass securetunneling component and does not enforce single process when MQTT is disabled.

### DIFF
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -315,15 +315,6 @@ int main(int argc, char *argv[])
         LOG_WARN(TAG, "Unable to append current working directory to PATH environment variable.");
     }
 
-    /**
-     * init() is currently responsible for making sure only 1 instance of Device Client is running at a given time.
-     * In the future, we may want to move other Device Client startup logic into this function.
-     * returns false if an exception is thrown
-     */
-    if (!init(argc, argv))
-    {
-        return 1;
-    }
     features = make_shared<FeatureRegistry>();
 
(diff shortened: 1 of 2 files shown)


xlcheng1 (approved):

shangabl (approved):

source/main.cpp
```diff
@@ -384,6 +384,15 @@ int main(int argc, char *argv[])
      * features.
      */
 #if !defined(DISABLE_MQTT)
+    /**
+     * init() is currently responsible for making sure only 1 instance of Device Client is running at a given time.
```
HarshGandhi-AWS, line 388: why do we have to initialize twice?

xlcheng1: the first initialization is under the condition `if (config.config.fleetProvisioning.enabled &&
        !config.config.fleetProvisioningRuntimeConfig.completedFleetProvisioning)`

source/main.cpp
```diff
@@ -354,6 +345,15 @@ int main(int argc, char *argv[])
          * Establish MQTT connection using claim certificates and private key to provision the device/thing.
          */
 #    if !defined(DISABLE_MQTT)
+        /**
+         * init() is currently responsible for making sure only 1 instance of Device Client is running at a given time.
```
HarshGandhi-AWS, line 349: I think the wording over here is confusing. It should be something like this.. 
```
init() is currently responsible for making sure only 1 MQTT connection is established on Device Client...
```
any thoughts @shangabl

PR #398, opened 2023-03-22
Fix failing amazonlinux integraton test image build

RogerZhongAWS:
### Motivation
- update curl library files to use latest version and build from source. 

### Testing
- verified aws-cpp-sdk and integration tests build successfully


HarshGandhi-AWS:
I feel this PR is no longer required since we are now using Ubuntu OS for testing. @RogerZhongAWS can you confirm and close this PR?

PR #408, opened 2023-05-02
Excluded feature validation and file creation if DISABLE_MQTT flag is enabled.

HarshGandhi-AWS:
Apart from Secure Tunneling feature excluded all other feature validation and file creation when `DISABLE_MQTT` flag is enabled.

### Motivation
When `DISABLE_MQTT` is enabled, only ST feature is built in the binary so it does not make sense to read other feature config and validate it. 

#### Revision diff summary
N/A

### Testing
Manually tested build

### DIFF
--- a/source/config/Config.cpp
+++ b/source/config/Config.cpp
@@ -276,11 +276,15 @@ bool PlainConfig::LoadFromCliArgs(const CliArgs &cliArgs)
         thingName = cliArgs.at(PlainConfig::CLI_THING_NAME).c_str();
     }
 
-    return logConfig.LoadFromCliArgs(cliArgs) && jobs.LoadFromCliArgs(cliArgs) && tunneling.LoadFromCliArgs(cliArgs) &&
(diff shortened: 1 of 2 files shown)


source/config/Config.cpp
```diff
@@ -2864,7 +2872,9 @@ bool Config::ParseConfigFile(const string &file, ConfigFileType configFileType)
         }
     }
 
+#if !defined(DISABLE_MQTT)
     LOGM_INFO(TAG, "Successfully fetched JSON config file: %s", Sanitize(contents).c_str());
```
RogerZhongAWS, line 2876: what is the reasoning behind checking the flag here?

RogerZhongAWS (approved):

source/config/Config.cpp
HarshGandhi-AWS: each line is printed as error message in GG component.

PR #415, opened 2023-05-27
Updated path to config file in aws-iot-device-client.service

dbouras:
This must have been a typo - can't think why any one would want to hide the configuration in a dot-directory under /etc

### Motivation
Noticed the issue while integrating aws-iot-device-client with a Yocto build

#### Change summary
A minor typo correction that impacts the location of the configuration file

### DIFF
--- a/setup/aws-iot-device-client.service
+++ b/setup/aws-iot-device-client.service
@@ -4,7 +4,7 @@ Wants=network-online.target
 After=network.target network-online.target
 
 [Service]
-Environment="CONF_PATH=/etc/.aws-iot-device-client/aws-iot-device-client.conf"
+Environment="CONF_PATH=/etc/aws-iot-device-client/aws-iot-device-client.conf"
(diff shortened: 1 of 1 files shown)


setup/aws-iot-device-client.service
```diff
@@ -4,7 +4,7 @@ Wants=network-online.target
 After=network.target network-online.target
 
 [Service]
-Environment="CONF_PATH=/etc/.aws-iot-device-client/aws-iot-device-client.conf"
```
HarshGandhi-AWS, line 7: The path is correct. We intentionally create device client resources in an hidden file 

https://github.com/awslabs/aws-iot-device-client/blob/f207eafc31b723492e33d5941fe1d05ba46c59b8/docs/SETUP.md?plain=1#L29

HarshGandhi-AWS:
Closing this PR since the path is right and no change will be required. Please feel free to open the PR if you disagree or feel we should update the path. Would be happy to discuss.

Issue #411, opened 2023-05-08
Fatal error condition occurred: aws-c-common string.c:217: allocator && aws_byte_cursor_is_valid(cursor)


thomas-roos:
**Describe the bug**

root@qemux86-64:/usr/lib/aws-iot-device-client/ptest# ./test-aws-iot-device-client 
[==========] Running 276 tests from 32 test suites.
[----------] Global test environment set-up.
[----------] 48 tests from ConfigTestFixture
[ RUN      ] ConfigTestFixture.AllFeaturesEnabled
Fatal error condition occurred in /usr/src/debug/aws-c-common/0.8.19-r0/git/source/string.c:217: allocator && aws_byte_cursor_is_valid(cursor)
Exiting Application
################################################################################
Stack trace:
################################################################################
/usr/lib/libaws-c-common.so.1(aws_backtrace_print+0x62) [0x7f974e575282]
/usr/lib/libaws-c-common.so.1(aws_fatal_assert+0x48) [0x7f974e55a428]
/usr/lib/libaws-c-common.so.1(+0x37599) [0x7f974e57d599]
/usr/lib/libaws-c-common.so.1(aws_json_value_get_from_object+0x29) [0x7f974e56cbf9]
/usr/lib/libaws-crt-cpp.so(_ZNK3Aws3Crt8JsonView11ValueExistsEPKc+0x1d) [0x7f974e6ef47d]
./test-aws-iot-device-client(+0x9b1b9) [0x555e6911d1b9]
./test-aws-iot-device-client(+0x11ce97) [0x555e6919ee97]
./test-aws-iot-device-client(+0x26223d) [0x555e692e423d]
./test-aws-iot-device-client(+0x2562e6) [0x555e692d82e6]
./test-aws-iot-device-client(+0x256572) [0x555e692d8572]
./test-aws-iot-device-client(+0x25671c) [0x555e692d871c]
./test-aws-iot-device-client(+0x256ed3) [0x555e692d8ed3]
./test-aws-iot-device-client(+0x26275d) [0x555e692e475d]
./test-aws-iot-device-client(+0x257051) [0x555e692d9051]
./test-aws-iot-device-client(+0x73d6f) [0x555e690f5d6f]
/lib/libc.so.6(__libc_start_main+0xeb) [0x7f974e0e3cfb]
./test-aws-iot-device-client(+0x8475a) [0x555e6910675a]
Aborted

**Environment (please complete the following information):**
[...]

PR #416, opened 2023-06-12
Updated SDK commit to use the latest SDK code

HarshGandhi-AWS:
### Motivation
- Issue number: #411 

#### Change summary
* Updated SDK commit to use the latest SDK code.
* Updated resource manager `initializeAllocator` method to public so it can be called from outside of `SharedCrtResourceManager`. 
* Added null check in `SharedCrtResourceManager::disconnect` method. 
* Updated unit test code to initialize allocator before using CRT library methods like CJSON or Crt::Map.
* Reverted this change to avoid calling `init` twice when FP is enabled. https://github.com/awslabs/aws-iot-device-client/pull/391/files

#### Revision diff summary
* Addressed reviewers comments 
* Added null check in `SharedCrtResourceManager::disconnect` method. 
* Updated unit test code to initialize allocator before using CRT library methods like CJSON or Crt::Map.

### Testing
- Manually tested changes locally 
- Ran CI tests 
- Ran Unit tests locally. 

### DIFF
--- a/source/SharedCrtResourceManager.cpp
+++ b/source/SharedCrtResourceManager.cpp
@@ -39,9 +39,29 @@ bool SharedCrtResourceManager::initialize(
     std::shared_ptr<Util::FeatureRegistry> featureRegistry)
 {
     features = featureRegistry;
-    initializeAllocator(config);
     initialized = buildClient(config) == SharedCrtResourceManager::SUCCESS;
     return initialized;
+}
+
+void SharedCrtResourceManager::loadMemTraceLevelFromEnvironment()
+{
+    const char *memTraceLevelStr = std::getenv("AWS_CRT_MEMORY_TRACING");
+    if (memTraceLevelStr)
+    {
+        switch (atoi(memTraceLevelStr))
+        {
+            case AWS_MEMTRACE_BYTES:
+                LOG_DEBUG(Config::TAG, "Set AWS_CRT_MEMORY_TRACING=AWS_MEMTRACE_BYTES");
+                memTraceLevel = AWS_MEMTRACE_BYTES;
+                break;
+            case AWS_MEMTRACE_STACKS:
+                LOG_DEBUG(Config::TAG, "Set AWS_CRT_MEMORY_TRACING=AWS_MEMTRACE_STACKS");
+                memTraceLevel = AWS_MEMTRACE_STACKS;
+                break;
+            default:
+                break;
+        }
+    }
 }
 
 bool SharedCrtResourceManager::locateCredentials(const PlainConfig &config) const
@@ -174,22 +194,24 @@ bool SharedCrtResourceManager::setupLogging(const PlainConfig &config) const
     return true;
 }
 
-void SharedCrtResourceManager::initializeAllocator(const PlainConfig &config)
+void SharedCrtResourceManager::initializeAllocator()
 {
+    loadMemTraceLevelFromEnvironment();
     allocator = aws_default_allocator();
-    memTraceLevel = config.memTraceLevel;
+
     if (memTraceLevel != AWS_MEMTRACE_NONE)
     {
         // If memTraceLevel == AWS_MEMTRACE_STACKS(2), then by default 8 frames per stack are used.
         allocator = aws_mem_tracer_new(allocator, nullptr, memTraceLevel, 0);
     }
+
+    // We MUST declare an instance of the ApiHandle to perform global initialization
+    // of the SDK libraries
+    apiHandle = unique_ptr<ApiHandle>(new ApiHandle());
 }
 
 int SharedCrtResourceManager::buildClient(const PlainConfig &config)
 {
-    // We MUST declare an instance of the ApiHandle to perform global initialization
-    // of the SDK libraries
-    apiHandle = unique_ptr<ApiHandle>(new ApiHandle());
     if (config.logConfig.sdkLoggingEnabled)
     {
         if (!setupLogging(config))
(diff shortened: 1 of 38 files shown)


source/main.cpp
```diff
@@ -289,6 +289,18 @@ namespace Aws
 
 int main(int argc, char *argv[])
 {
+    auto listener = std::make_shared<DefaultClientBaseNotifier>();
```
marcoemorais-aws, line 292: Currently `SharedCrtResourceManager::initialize` does 3 things.
1. Stores a reference to the feature registry 
2. Initializes the allocator (`SharedCrtResourceManager::initializeAllocator`)
3. Initializes the SDK client handle (`SharedCrtResourceManager::buildClient`)

Step 3 requires that the `Config` passed to `SharedCrtResourceManager::initialize` be initialized with the `CliArgs` **before** calling this function whereas the proposed change on this commit initializes `Config`  from `CliArgs` **after** calling `SharedCrtResourceManager::initialize`.

In order to avoid breaking this invariant, I recommend that you make the following changes to this commit.

* Make the function to initialize the allocator public (`SharedCrtResourceManager::initializeAllocator`) and remove the `Config` from the signature since it actually isn't used to initialize the allocator. 
* Remove the call to `SharedCrtResourceManager::initializeAllocator` from `SharedCrtResourceManager::initialize`.
* Create the shared instance to `resourceManager` as the first operation in `main` and invoke the `SharedCrtResourceManager::initializeAllocator`.

The other changes in this commit aren't required.

marcoemorais-aws (dismissed):

RogerZhongAWS (commented):
You also may want to check if this line can be pulled out of buildClient()

https://github.com/awslabs/aws-iot-device-client/blob/239d10d968eb635d883998ff70e5054e66aef450/source/SharedCrtResourceManager.cpp#LL192C56-L192C56

test/config/TestConfigSerialization.cpp
```diff
@@ -138,45 +123,14 @@ INSTANTIATE_TEST_CASE_P(
             }
         ]
     }
-})",
-        R"({
```
HarshGandhi-AWS, line 142: I have removed this tests since it looks redundant.

source/config/Config.cpp
```diff
@@ -287,7 +287,7 @@ bool PlainConfig::LoadFromCliArgs(const CliArgs &cliArgs)
     return loadFeatureCliArgs;
 }
 
-bool PlainConfig::LoadFromEnvironment()
+void PlainConfig::LoadMemTraceLevelFromEnvironment()
```
marcoemorais-aws, line 290: @HarshGandhi-AWS This solution will work, but what if we completely remove the initialization order dependency between Config and SharedCrtResourceManager by doing the following 2 steps? 

1. Copy-paste the code from `PlainConfig::LoadMemTraceLevelFromEnvironment()` to `SharedCrtResourceManager::initializeAllocator`.
2. Invoke `SharedCrtResourceManager::initializeAllocator`  from the `SharedCrtResourceManager` constructor. 

This has the following benefits:
* Config does not have to be initialized before SharedCrtResourceManager.
* Unit tests that use SharedCrtResourceManager do not require a fixture to explicitly initialize SharedCrtResourceManager for memory tracing.
* main.cpp requires only a one-line change to invoke the `resourceManager = std::make_shared<SharedCrtResourceManager>();`

source/SharedCrtResourceManager.cpp
```diff
@@ -174,22 +173,22 @@ bool SharedCrtResourceManager::setupLogging(const PlainConfig &config) const
     return true;
 }
 
-void SharedCrtResourceManager::initializeAllocator(const PlainConfig &config)
+void SharedCrtResourceManager::initializeAllocator(const aws_mem_trace_level &configMemTraceLevel)
 {
-    allocator = aws_default_allocator();
```
marcoemorais-aws, line 179: The memory trace allocator **wraps** the allocator being traced and as a result I think you must keep this line.

```
allocator = aws_default_allocator();
```

test/config/TestConfigSerialization.cpp
marcoemorais-aws: I think these tests are roundtrip tests to verify the string used as input to step 1 is the same as the string returned as output in step 2.
1. read config from a string into json object using `LoadFromJson`
2. return string from in-memory json object using `SerializeToObject`

test/config/TestConfig.cpp
```diff
@@ -154,6 +156,8 @@ TEST_F(ConfigTestFixture, AllFeaturesEnabled)
         "secure-element-token-label": "token-label"
       }
 })";
+    SharedCrtResourceManager resourceManager;
```
marcoemorais-aws, line 159: Isn't one of these already initialized in `ConfigTestFixture`?

test/config/TestConfigSerialization.cpp
HarshGandhi-AWS: Okay. I have separated the older test into two different tests.

source/config/Config.cpp
HarshGandhi-AWS: Calling the initializeAllocator() method from SharedCrtResourceManager constructor was causing some Sensor Publish feature unit test failures. To avoid rewriting the unit tests, I have removed initializeAllocator() method call from the constructor and I am manually calling the initializeAllocator() method from main and some unit tests wherever the Cjson library was used.

RogerZhongAWS (approved):

RogerZhongAWS (dismissed):

RogerZhongAWS (approved):

BolongZhang-AWS (dismissed):

BolongZhang-AWS (approved):
ship

PR #426, opened 2023-09-28
Updated code to shutdown on tunnel close callback

HarshGandhi-AWS:
### Motivation
- Greengrass Secure Tunneling component which use Device Client code was not closing the thread on getting the tunnel close notification.

#### Change summary
- Updated code to shutdown Device Client on tunnel close callback

#### Revision diff summary

### Testing
- Locally tested the changes does not affect the device client application logic


RogerZhongAWS (commented):
looks good, just may want to provide a little more context in the comment explaining why we have to do this hack.

RogerZhongAWS (approved):

PR #427, opened 2023-10-03
Updated Secure Tunneling close logic for Secure Tunneling Component

HarshGandhi-AWS:
### Motivation
- Secure tunneling component was failing to clean up the threads opened for each new tunnel. To resolve the issue, we are closing Device Client with `SIGTERM` signal. This will make sure that on tunnel close signal, we will send process termination signal which will eventually close all threads and free up all memory to avoid memory leak.

#### Change summary
- Calling signal termination from tunnel close callback 
- Updated abort function to pass in exit code. 

#### Revision diff summary

### Testing
- Tested locally by creating private component. 

### DIFF
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -157,7 +157,6 @@ void shutdown()
     }
 #endif
     LoggerFactory::getLoggerInstance()->shutdown();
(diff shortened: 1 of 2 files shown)


RogerZhongAWS (approved):
question addressed offline, LGTM

PR #473, opened 2024-11-11
Add job document comparator to ignore pre-signed difference in s3 url

ig15:
### Motivation
Currently 2 job documents containing the same s3 pre-signed url that are sent from the console are compared directly in string formatting. After these changes the 2 job documents will be compared neglecting the aws temporary access signature, to avoid 2 identical job documents with different signatures being considered different and creating conflicts on the device processes

#### Change summary
Added a comparator function to replace the pre-signed part of an s3 pre-signed url with a common placeholder to avoid
same job with different pre-signed parts to be inferred differently by device client

#### Revision diff summary

### Testing

Changes tested locally by simulating different cases for the job documents. 
Manually tested the output against customer job document and handler files as well.

### DIFF
--- a/source/jobs/JobsFeature.cpp
+++ b/source/jobs/JobsFeature.cpp
@@ -527,9 +527,87 @@ void JobsFeature::copyJobsNotification(Iotjobs::JobExecutionData job)
     latestJobsNotification.ExecutionNumber = job.ExecutionNumber.value();
 }
 
+
+bool JobsFeature::compareJobDocuments(const Aws::Crt::JsonObject& job1, const Aws::Crt::JsonObject& job2) {
+    std::string str1 = job1.View().WriteCompact().c_str();
+    std::string str2 = job2.View().WriteCompact().c_str();
+
+    // Regular expression to match S3 URLs and capture the non-presigned parts
+    std::regex s3UrlRegex(R"((https://[^.\s"]+\.s3[.-](?:[^.\s"]+\.)?amazonaws\.com/[^?\s"]+)(\?[^"\s]+)?)");
+
+    // Function to replace only the pre-signed portion of S3 URLs
+    auto processPresignedUrls = [&s3UrlRegex](std::string& s) -> int {
+        int count = 0;
+        std::string result;
+        std::sregex_iterator it(s.begin(), s.end(), s3UrlRegex);
+        std::sregex_iterator end;
+
+        size_t lastPos = 0;
+        for (; it != end; ++it) {
+            count++;
+            result += s.substr(lastPos, it->position() - lastPos);
+            result += it->str(1);
+            // Add the non-presigned part of the URL
+            result += "?aws:iot:s3-presigned-suffix:PLACEHOLDER"; 
+            // Replace only the presigned portion
+            lastPos = it->position() + it->length();
+        }
+        result += s.substr(lastPos);
+
+        s = result;
+        return count;
+    };
+
+    int count1 = processPresignedUrls(str1);
+    int count2 = processPresignedUrls(str2);
+
+    // If the number of pre-signed URLs differs, the documents are different
+    if (count1 != count2) {
+        return false;
+    }
+
+    std::cout << "Processed str1: " << str1 << std::endl;
+    std::cout << "Processed str2: " << str2 << std::endl;
+
+    return str1 == str2;
+}
 bool JobsFeature::isDuplicateNotification(JobExecutionData job)
 {
     unique_lock<mutex> readLatestNotificationLock(latestJobsNotificationLock);
+    
+    // Basic logging at the start of the function
+    std::cout << "DEBUG: Entering isDuplicateNotification function" << std::endl;
+
+
+
+    // Log Job IDs being compared
+    std::cout << "INFO: Comparing Job IDs - New: " << (job.JobId.has_value() ? job.JobId.value().c_str() : "NULL")
+              << ", Latest: " << (latestJobsNotification.JobId.has_value() ? latestJobsNotification.JobId.value().c_str() : "NULL") << std::endl;
+
+
+
+
+
+    // Log Job Documents being compared
+    std::cout << "INFO: Comparing Job Documents - New: " 
+              << (job.JobDocument.has_value() ? job.JobDocument.value().View().WriteCompact().c_str() : "NULL")
+              << ", Latest: " 
+              << (latestJobsNotification.JobDocument.has_value() ? latestJobsNotification.JobDocument.value().View().WriteCompact().c_str() : "NULL") 
+              << std::endl;
+
+
+
+
+
+        // Log Execution Numbers being compared
+    std::cout << "INFO: Comparing Execution Numbers - New: " 
+              << (job.ExecutionNumber.has_value() ? std::to_string(job.ExecutionNumber.value()) : "NULL")
+              << ", Latest: " 
+              << (latestJobsNotification.ExecutionNumber.has_value() ? std::to_string(latestJobsNotification.ExecutionNumber.value()) : "NULL") 
+              << std::endl;
+
+
+
     if (!latestJobsNotification.JobId.has_value())
     {
         // We have not seen a job yet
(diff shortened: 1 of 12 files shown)


source/jobs/JobsFeature.h
```diff
@@ -313,6 +317,17 @@ namespace Aws
...
+                    /**
+                     * \brief Compares two job documents, ignoring differences in pre-signed URLs.
+                     *
+                     * @param job1 The first job document as a JsonObject
+                     * @param job2 The second job document as a JsonObject
+                     * @return true if the documents are equivalent (ignoring pre-signed URLs), false otherwise
+                     */
+                        
+                    // Add this line to declare the test class as a friend
+                    friend class TestJobsFeaturePrivate;
```
tony-josi-aws, line 329: Ideally, its better to have no information about the tests in the code we are testing.  

Since `compareJobDocuments` is a utility function that doesn't need access (or be related) to any of the other members of the `JobsFeature` class, it might be a better idea to move this function to a file in the [util folder](https://github.com/awslabs/aws-iot-device-client/tree/main/source/util).

It also helps in testing the `compareJobDocuments` individually, similar to other APIs in the util folder without the need to have a friend class. The mock of the `compareJobDocuments` can be used in the test for `isDuplicateNotification`.

source/jobs/JobsFeature.h
```diff
@@ -77,6 +78,9 @@ namespace Aws
                     virtual int stop() override;
 
                   protected:
+
+                    bool compareJobDocuments(const Aws::Crt::JsonObject& job1, const Aws::Crt::JsonObject& job2);
```
tony-josi-aws, line 82: Missing API documentation (maybe add some info about how the comparrision is performed).

test/jobs/TestJobsFeature.cpp
```diff
@@ -710,3 +730,39 @@ TEST_F(TestJobsFeature, InvalidJobDocument)
...
+    Aws::Crt::JsonObject doc5("{\"url\": \"https://bucket.s3.amazonaws.com/file?AWSAccessKeyId=AKIAIOSFODNN7EXAMPLE&Expires=9876543210&Signature=YYYYYYYYYYYYYYYYYYYYYYYYYYY\"}");
+    EXPECT_TRUE(compareJobDocuments(doc4, doc5));
+
+    // Test case 4: Documents with multiple pre-signed URLs
+    Aws::Crt::JsonObject doc6("{\"url1\": \"https://bucket1.s3.amazonaws.com/file1?AWSAccessKeyId=AKIAIOSFODNN7EXAMPLE&Expires=1234567890&Signature=XXXXXXXXXXXXXXXXXXXXXXXXXXX\", \"url2\": \"https://bucket2.s3.amazonaws.com/file2?AWSAccessKeyId=AKIAIOSFODNN7EXAMPLE&Expires=1234567890&Signature=ZZZZZZZZZZZZZZZZZZZZZZZZZZZ\"}");
+    Aws::Crt::JsonObject doc7("{\"url1\": \"https://bucket1.s3.amazonaws.com/file1?AWSAccessKeyId=AKIAIOSFODNN7EXAMPLE&Expires=9876543210&Signature=YYYYYYYYYYYYYYYYYYYYYYYYYYY\", \"url2\": \"https://bucket2.s3.amazonaws.com/file2?AWSAccessKeyId=AKIAIOSFODNN7EXAMPLE&Expires=9876543210&Signature=WWWWWWWWWWWWWWWWWWWWWWWWWWW\"}");
+    EXPECT_TRUE(compareJobDocuments(doc6, doc7));
+
+    // Test case 6: Documents with multiple pre-signed URLs in one doc and single url in the other
```
tony-josi-aws, line 755: Are we testing the scenario where there are multiple pre-signed URLs in both documents (which may or may not be the same URLs in both documents), but the order is different?

For example:

```
    Aws::Crt::JsonObject doc6("{\"url1\": \"https://bucket1.s3.amazonaws.com/file1?AWSAccessKeyId=AKIAIOSFODNN7EXAMPLE&Expires=1234567890&Signature=XXXXXXXXXXXXXXXXXXXXXXXXXXX\", \"url2\": \"https://bucket2.s3.amazonaws.com/file2?AWSAccessKeyId=AKIAIOSFODNN7EXAMPLE&Expires=1234567890&Signature=ZZZZZZZZZZZZZZZZZZZZZZZZZZZ\"}");
    Aws::Crt::JsonObject doc7("{\"url1\": \"https://bucket2.s3.amazonaws.com/file2?AWSAccessKeyId=AKIAIOSFODNN7EXAMPLE&Expires=9876543210&Signature=WWWWWWWWWWWWWWWWWWWWWWWWWWW\", \"url2\": \"https://bucket1.s3.amazonaws.com/file1?AWSAccessKeyId=AKIAIOSFODNN7EXAMPLE&Expires=9876543210&Signature=YYYYYYYYYYYYYYYYYYYYYYYYYYY\"}");
```

source/jobs/JobsFeature.cpp
```diff
@@ -527,9 +527,87 @@ void JobsFeature::copyJobsNotification(Iotjobs::JobExecutionData job)
...
+
+    // If the number of pre-signed URLs differs, the documents are different
+    if (count1 != count2) {
+        return false;
+    }
+
+    std::cout << "Processed str1: " << str1 << std::endl;
+    std::cout << "Processed str2: " << str2 << std::endl;
+
+    return str1 == str2;
```
tony-josi-aws, line 572: Does this work if there is the same set of URLs in both documents but the order is different in each document?

tony-josi-aws (changes requested):

source/jobs/JobsFeature.cpp
```diff
@@ -527,9 +527,87 @@ void JobsFeature::copyJobsNotification(Iotjobs::JobExecutionData job)
...
+
+    int count1 = processPresignedUrls(str1);
+    int count2 = processPresignedUrls(str2);
+
+    // If the number of pre-signed URLs differs, the documents are different
+    if (count1 != count2) {
+        return false;
+    }
+
+    std::cout << "Processed str1: " << str1 << std::endl;
```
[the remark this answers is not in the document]
ig15: These couts were only added for debugging and will be removed in the revised PR.

source/jobs/JobsFeature.cpp
```diff
@@ -527,9 +527,87 @@ void JobsFeature::copyJobsNotification(Iotjobs::JobExecutionData job)
...
+    std::cout << "Processed str2: " << str2 << std::endl;
+
+    return str1 == str2;
+}
 bool JobsFeature::isDuplicateNotification(JobExecutionData job)
 {
     unique_lock<mutex> readLatestNotificationLock(latestJobsNotificationLock);
+    
+    // Basic logging at the start of the function
+    std::cout << "DEBUG: Entering isDuplicateNotification function" << std::endl;
```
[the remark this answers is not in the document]
ig15: This will also be removed.

source/jobs/JobsFeature.h
ig15: Modifications will be made in the revised PR.

source/jobs/JobsFeature.h
ig15: Will be taken care of in the revision.

test/jobs/TestJobsFeature.cpp
ig15: Considering that the job documents are sent by IOT core due to the subscription mechanisms, it is not possible in theory for the order of the 2 job documents to differ which has been confirmed via manual testing as well. Still, if they differ only in the order of the URLs, it is most likely cause of an actual separate job sent and hence they should be inferred as a different jobs altogether.

source/jobs/JobsFeature.cpp
ig15: As discussed in previous comments, if the order is different then the jobs are to be considered as separate jobs.

amazonKamath (approved):

amazonKamath (approved):

tony-josi-aws (approved):

source/jobs/JobsFeature.cpp
```diff
@@ -527,9 +527,50 @@ void JobsFeature::copyJobsNotification(Iotjobs::JobExecutionData job)
...
+        std::string result;
+        std::sregex_iterator it(s.begin(), s.end(), s3UrlRegex);
+        std::sregex_iterator end;
+
+        size_t lastPos = 0;
+        for (; it != end; ++it) {
+            count++;
+            result += s.substr(lastPos, it->position() - lastPos);
+            result += it->str(1);
+            // Add the non-presigned part of the URL
```
RogerZhongAWS, line 550: code overall looks good to me. one small nitpick is to leave some whitespace around lines of code that have a comment associated with them so it is easier to tell which goes with which

RogerZhongAWS (approved):
code overall looks good to me. one small nitpick is to leave some whitespace around lines of code that have a comment associated with them so it is easier to tell which goes with which

source/jobs/JobsFeature.cpp
ig15: Will make the changes.

Issue #488, opened 2025-02-12
Non-persistent config shadow changes


AleksandarGT:
I have enabled config shadows in the configuration because we'd like to change the configuration on demand from the cloud directly. However, the changed configuration is not persistent. After changing the config shadow, if I restart the aws device client service the changes are applied but are not saved in the actual configuration file, so on next service restart / device reboot the changes will revert back to the real configuration file.

It's important to note that we have registered the device client as our own service rather than the automatic way through the setup cli because we had problems, I am not sure if this has anything to do with the current issue.

Our own service file:

[Unit]
Description=AWS IoT Device Client
Wants=network-online.target
After=network.target network-online.target

[Service]
User=sens
Type=simple
Environment="CONF_PATH=/home/sens/.aws-iot-device-client/aws-iot-device-client.conf"
ExecStart=/home/sens/aws-iot-core-connection/aws-iot-device-client/build/aws-iot-device-client --config-file $CONF_PATH
StandardOutput=append:/home/sens/aws-iot-core-connection/aws-iot-device-client/logs/system.log
StandardError=append:/home/sens/aws-iot-core-connection/aws-iot-device-client/logs/system-error.log
Restart=always
RestartSec=5

[Install]
WantedBy=multi-user.target

ig15:
Hey @AleksandarGT. Can you build with the changes in this PR to confirm if it fixes the problem you're facing: https://github.com/awslabs/aws-iot-device-client/pull/490

AleksandarGT:
The changes to the shadow via aws now reflects to the configuration file. 

However, when making a change to the shadow via the cloud, the config changes for the jobs get includeStdOut omitted. This happens even if I change another configuration option not for the jobs.

Before

      "jobs":	{
        "enabled":	true,
        "includeStdOut": true,
        "handler-directory": "/home/sens/.aws-iot-device-client/jobs"
      },

after

      "jobs":	{
        "enabled":	true,
        "handler-directory": "/home/sens/.aws-iot-device-client/jobs"
      },

Also we use fleet provisioning, so when the initial provisioning is complete a file is created .aws-iot-device-client/aws-iot-device-client-runtime.conf

When I make a change to the config shadow via the cloud, the information from there gets copied over to the aws-iot-device-client.conf. The whole 'runtime-config' json wasn't a part of the configuration before the shadow update. I am not sure this is desired behavior.

	"fleet-provisioning":	{
		"enabled":	true,
[...]

PR #490, opened 2025-02-21
Update Config Shadow to store config file updates persistently

ig15:
### Motivation
- Issue number: https://github.com/awslabs/aws-iot-device-client/issues/488

#### Change summary
The function added modifies the current config shadow feature behavior to update the local configuration files of device client with the new configurations sent via remote config shadow so that they persist over multiple device client restarts

### Testing
Tested the changes locally with multiple config shadow updates

### DIFF
--- a/source/config/Config.h
+++ b/source/config/Config.h
@@ -486,6 +486,7 @@ namespace Aws
                 static constexpr char DEFAULT_CONFIG_DIR[] = "~/.aws-iot-device-client/";
                 static constexpr char DEFAULT_KEY_DIR[] = "~/.aws-iot-device-client/keys/";
                 static constexpr char DEFAULT_CONFIG_FILE[] = "~/.aws-iot-device-client/aws-iot-device-client.conf";
+                static constexpr char DEFAULT_ROOT_CONFIG_FILE[] = "/etc/.aws-iot-device-client/aws-iot-device-client.conf";
                 static constexpr char DEFAULT_FLEET_PROVISIONING_RUNTIME_CONFIG_FILE[] =
                     "~/.aws-iot-device-client/aws-iot-device-client-runtime.conf";
                 static constexpr char DEFAULT_HTTP_PROXY_CONFIG_FILE[] = "~/.aws-iot-device-client/http-proxy.conf";
--- a/source/config/Config.h
+++ b/source/config/Config.h
@@ -488,7 +488,7 @@ namespace Aws
                 static constexpr char DEFAULT_CONFIG_DIR[] = "~/.aws-iot-device-client/";
                 static constexpr char DEFAULT_KEY_DIR[] = "~/.aws-iot-device-client/keys/";
                 static constexpr char DEFAULT_CONFIG_FILE[] = "~/.aws-iot-device-client/aws-iot-device-client.conf";
-                static constexpr char DEFAULT_ROOT_CONFIG_FILE[] = "/etc/.aws-iot-device-client/aws-iot-device-client.conf";
+                static constexpr char DEFAULT_SYSTEM_CONFIG_FILE[] = "/etc/.aws-iot-device-client/aws-iot-device-client.conf";
                 static constexpr char DEFAULT_FLEET_PROVISIONING_RUNTIME_CONFIG_FILE[] =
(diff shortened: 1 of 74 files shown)


source/shadow/ConfigShadow.cpp
```diff
@@ -19,10 +19,31 @@ using namespace Aws::Iotshadow;
...
+
+    if (configFile.is_open())
+    {
+        configFile << jsonObj.View().WriteReadable();
+        configFile.close();
+        LOGM_INFO(TAG, "Updated configuration saved locally to disk at: %s", configFilePath);
+    }
+    else
+    {
+    LOGM_WARN(TAG, "Failed to open config file: %s, Error: %s", configFilePath, strerror(errno));
```
tony-josi-aws, line 43: NIT: Formatting

source/config/Config.h
```diff
@@ -486,6 +488,7 @@ namespace Aws
                 static constexpr char DEFAULT_CONFIG_DIR[] = "~/.aws-iot-device-client/";
                 static constexpr char DEFAULT_KEY_DIR[] = "~/.aws-iot-device-client/keys/";
                 static constexpr char DEFAULT_CONFIG_FILE[] = "~/.aws-iot-device-client/aws-iot-device-client.conf";
+                static constexpr char DEFAULT_ROOT_CONFIG_FILE[] = "/etc/.aws-iot-device-client/aws-iot-device-client.conf";
```
tony-josi-aws, line 491: Is this path fixed for all platforms where device client is supposed to run.

source/shadow/ConfigShadow.h
```diff
@@ -200,6 +200,14 @@ namespace Aws
                      * @param jsonObj contains all device client features' configuration
                      */
                     void loadFeatureConfigIntoJsonObject(PlainConfig &config, Aws::Crt::JsonObject &jsonObj) const;
+                    /**
+                     * \brief Updates the local configuration file of device client with the config shadow update 
+                     * received from the AWS IOT core for persistence of the new device configuration 
+                     * @param config device client local configuration
+                     * @param configFilePath file path of the local aws-iot-device-client.conf
```
tony-josi-aws, line 207: Is it more correct to say, Absolute file path of device client configuration?

source/shadow/ConfigShadow.cpp
```diff
@@ -352,6 +373,38 @@ void ConfigShadow::resetClientConfigWithJSON(
...
+    if(config.configShadow.persistentUpdate)
+    {
+        JsonObject jsonObj;
+        config.SerializeToObject(jsonObj);  
+        const char* jsonStr = jsonObj.View().WriteReadable().c_str();
+        LOGM_INFO(TAG, "Updating device configuration files with the following config shadow update: %s", jsonStr);
+        
+        string userConfig = FileUtils::ExtractExpandedPath(Config::DEFAULT_CONFIG_FILE);
+        updateLocalConfigFile(config, userConfig.c_str());
+        updateLocalConfigFile(config, Config::DEFAULT_ROOT_CONFIG_FILE);
```
DakshitBabbar, line 406: Just a doubt: why is `Config::DEFAULT_ROOT_CONFIG_FILE` passed to the `updateLocalConfigFile` function directly while `Config::DEFAULT_CONFIG_FILE` is first parsed using the FileUtils method?

ig15: The updateLocalConfigFile function is not able to parse DEFAULT_CONFIG_FILE directly since it is the shorthand home directory filepath  i.e. ~/ unlike DEFAULT_ROOT_CONFIG_FILE which is the absolute filepath from the root i.e. /
Hence we need to expand that filepath to its absolute path before passing it to the function.

source/shadow/ConfigShadow.h
ig15: Yes. Will update it. Thanks

source/shadow/ConfigShadow.cpp
ig15: Will fix in the revision

DakshitBabbar (approved):

tony-josi-aws (approved):

PR #515, opened 2025-05-08
Add ubuntu 18.04 base image build for Secure Tunneling executable build

ig15:
### Motivation

Ubuntu 22.04 compiles the ST executable with newer glibc versions which makes the executable incompatible with target systems with older glibc versions support due to dynamic linker errors. 
GLIBC is not forward-compatible. A system with glibc 2.17 does not provide 2.25 symbols.
Hence, we are adding ubuntu 18.04 base image build for device client for the ST executable to be compiled on.

#### Change summary

Ubuntu 18.04 base image Dockerfile and workflow job added.

#### Revision diff summary

### Testing
Changes tested on feature branch

### DIFF
--- a/.github/docker-images/base-images/device-client/ubuntu-18.04/Dockerfile
+++ b/.github/docker-images/base-images/device-client/ubuntu-18.04/Dockerfile
@@ -17,6 +17,7 @@ RUN apt-get update -qq \
     libc6-dbg \
     softhsm \
     zlib1g-dev \
+    libssl-dev \
     && apt-get clean
 
 ###############################################################################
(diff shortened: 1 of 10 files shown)


tony-josi-aws (approved):

.github/docker-images/base-images/device-client/ubuntu-18.04/Dockerfile
```diff
@@ -0,0 +1,120 @@
+FROM ubuntu:18.04 AS base
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+ARG OPENSSL_VERSION=3.0.8
```
amazonKamath, line 5: is this the version used in 22.04 as well?

ig15: Yes

